PR #1, opened 2021-04-23
Ensure display library tested

kenoir:
In the switch to catalogue-api repo, the CI tests for display were missed, this re-adds them.

Part of https://github.com/wellcomecollection/catalogue/issues/1572


.buildkite/pipeline.yml
```diff
@@ -2,6 +2,8 @@ steps:
   - command: .buildkite/scripts/run_autoformat.py
     label: "autoformat"
   - wait
+  - command: .buildkite/scripts/run_job.py display
+    label: "display"
```
alexwlchan, line 6: In the catalogue repo we had a wait between the libraries and the applications – if display is broken, the applications definitely will be.

I can imagine we might start extracting other builds into common libraries (e.g. HTTP-related code) that would benefit from similar "stages" in CI.

kenoir: Makes sense. Are you ok if I merge this - then follow up with splitting it into a another stage (I think also there is some stacks common tests that I will also put there).

alexwlchan (approved):
Yup, good to merge as-is.

PR #6, opened 2021-04-27
Add search namespace

kenoir:
This change creates a `uk.ac.wellcome.platform.api.search` namespace and moves the previous `api` service into it.

The stacks & snapshot services will follow in a future PR.

This change is to enable https://github.com/wellcomecollection/wellcomecollection.org/issues/6337


jamesgorrie (approved):
✨

PR #9, opened 2021-04-27
Remove the Alpakka dependency from the snapshot generator

alexwlchan:
This fell out of the work in https://github.com/wellcomecollection/platform/issues/5101

The snapshot generator had a dependency on the Alpakka library, which has historically been a source of pain whenever you touch the Akka version. Akka is extremely picky about versions, and if you ever had mismatched versions in the classpath, snapshot generator tests would start failing. Sorting this is an incredibly frustrating process, one which I relived this morning.

We only use a tiny bit of the Alpakka functionality – for multi-part uploads to S3 – and that's simple enough that I figured I could rewrite it.

This PR rewrites the snapshot generator to be sans-Akka, and in particular sans-Alpakka. The code is also 350 lines lighter, has a few more tests, and should be much easier to debug on the next go-round. Note that the feature tests are unchanged, modulo imports – this is just an internal refactor. None of the external interfaces have changed.


kenoir (approved):
You rewrote the multi-part upload! 🔥

PR #12, opened 2021-04-27
New API stacks

jamieparkinson:
This creates new API gateways that will be found at 
- https://catalogue.api-stage-delta.wellcomecollection.org
- https://catalogue.api-prod.wellcomecollection.org

Only the stage stack is applied currently (eg https://catalogue.api-stage-delta.wellcomecollection.org/catalogue/v2/works). After these are both up we will look into shifting across the cloudfront origins, then removing the old stacks and flattening the terraform directory structure. Most of the code here is copied from those stacks.


terraform/stack/alarm.tf
```diff
@@ -0,0 +1,23 @@
+//TODO does this do anything
```
kenoir, line 1: No. I think it's been a long time since it did. Arguably it should do something, but we're not going to tackle it here and might choose to do it differently if we did, so delete it.

terraform/catalogue_api/terraform.tf
```diff
@@ -0,0 +1,47 @@
...
+  config = {
+    role_arn = "arn:aws:iam::760097843905:role/platform-read_only"
+
+    bucket = "wellcomecollection-platform-infra"
+    key    = "terraform/platform-infrastructure/shared.tfstate"
+    region = "eu-west-1"
+  }
+}
+
+// TODO remove this, hold the resources in this module
```
kenoir, line 36: You can also hold the state here by importing it - as long a we remember to remove the state rather than delete the ECR repositories in the other stack.

terraform/catalogue_api/locals.tf
```diff
@@ -0,0 +1,12 @@
+locals {
+  catalogue_vpcs = data.terraform_remote_state.accounts_catalogue.outputs
+
+  vpc_id          = local.catalogue_vpcs["catalogue_vpc_id"]
+  private_subnets = local.catalogue_vpcs["catalogue_vpc_private_subnets"]
+
+  elastic_cloud_vpce_sg_id = data.terraform_remote_state.infra_critical.outputs["ec_catalogue_privatelink_sg_id"]
+
+  // TODO move this tf state into this module
+  search_api_repository = data.terraform_remote_state.catalogue_api_shared_old.outputs["ecr_api_repository_url"]
```
kenoir, line 10: this has become `ecr_search_repository_url`

terraform/catalogue_api/terraform.tf
jamieparkinson: This is (I think) what I meant by this comment, I was thinking I'd shift all the remaining state over when I come to delete the old stack

terraform/catalogue_api/cluster.tf
```diff
@@ -0,0 +1,3 @@
+resource "aws_ecs_cluster" "catalogue_api" {
+  name = "catalogue-api-20210426"
```
alexwlchan, line 2: Is it too late to add dashes to the date? I find `2021-04-26` much easier to read than `20210426`.

terraform/catalogue_api/main.tf
```diff
@@ -0,0 +1,45 @@
...
+
+  vpc_id                   = local.vpc_id
+  private_subnets          = local.private_subnets
+  elastic_cloud_vpce_sg_id = local.elastic_cloud_vpce_sg_id
+  cluster_arn              = aws_ecs_cluster.catalogue_api.arn
+
+  providers = {
+    aws.dns = aws.dns
+  }
+}
```
alexwlchan, line 22: Do we want to have prod/stage in the same configuration? Historically we had them separate, so you could test changes in staging without blowing up prod.

terraform/catalogue_api/provider.tf
```diff
@@ -0,0 +1,27 @@
...
+
+provider "aws" {
+  alias = "platform"
+
+  region = var.aws_region
+
+  assume_role {
+    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
+  }
+}
```
alexwlchan, line 17: Are you using this provider anywhere?

terraform/catalogue_api/terraform.tf
```diff
@@ -0,0 +1,47 @@
+terraform {
+  backend "s3" {
+    role_arn = "arn:aws:iam::756629837203:role/catalogue-developer"
+
+    bucket         = "wellcomecollection-catalogue-infra-delta"
+    key            = "terraform/catalogue/api/catalogue_api.tfstate"
```
alexwlchan, line 6: Minor: the "catalogue" repo is no more, let's update this key to match:

```suggestion
    key            = "terraform/catalogue-api/catalogue_api.tfstate"
```

terraform/catalogue_api/terraform.tf
kenoir: I meant you could do it now, and then deal with the deleting the duplicated state from the other stack later - but either way is fine.

terraform/stack/security_groups.tf
```diff
@@ -0,0 +1,54 @@
...
+    from_port = 0
+    to_port   = 0
+    protocol  = "-1"
+    self      = true
+  }
+
+  tags = {
+    Name = "${var.environment_name}-catalogue_api-interservice"
+  }
+}
```
alexwlchan, line 54: Do we still need this? We used to have this so applications could talk to a logstash service, but I don't think it's needed any more

terraform/catalogue_api/cluster.tf
jamieparkinson: Nope not too late, will do

terraform/catalogue_api/main.tf
jamieparkinson: I think - as per some discussion with @kenoir in Slack - we do want them to be shared like this as it is more consistent with our other services, and it's easier to keep them in sync. You can still test them separately (as I have been doing) with a `terraform apply -target=module.catalogue_api_stage`

alexwlchan: \*nod* I must have missed that discussion.

Did you consider workspaces? https://www.terraform.io/docs/language/state/workspaces.html I haven't used them, but Digirati used them for the identity API and they sound interesting.

terraform/stack/security_groups.tf
jamieparkinson: Interesting question, let's try

kenoir (approved):

terraform/stack/security_groups.tf
kenoir: I think we do not. None of these services talk to their siblings.

terraform/catalogue_api/main.tf
kenoir: I don't think we want to use workspaces.

From the docs:

```
In particular, organizations commonly want to create a strong separation between multiple deployments of the same infrastructure serving different development stages (e.g. staging vs. production) or different internal teams. In this case, the backend used for each deployment often belongs to that deployment, with different credentials and access controls. Named workspaces are not a suitable isolation mechanism for this scenario.
```

jamieparkinson: I did not - it seems like they'd be very much equivalent to what they're doing albeit a bit more painful to configure differences (https://www.terraform.io/docs/language/state/workspaces.html#current-workspace-interpolation). I'll probably leave them out for now because I'm not sure how shifting some of the old state across would work with them (aka, better the devil you know)

jamieparkinson: @kenoir I was going to quote that but then I realised that holding the stacks in the same configuration suffers the exact same drawback 🤷‍♂️

terraform/catalogue_api/terraform.tf
jamieparkinson: Very neat that terraform can tell I've done this change and offers to copy across the state from the previous backend

Issue #8, opened 2021-04-27
Query Elasticsearch in Items API


kenoir:
In order to make the items service more performant and bring it in line with the search service, we should query elasticsearch for items information.

PR #28, opened 2021-04-30
Add Tracing, CustomDirectives to items-api

kenoir:
Bring the items-api more in line with search.

Towards https://github.com/wellcomecollection/catalogue-api/issues/8


items/src/main/scala/uk/ac/wellcome/platform/api/items/Main.scala
```diff
@@ -25,9 +27,30 @@ object Main extends WellcomeTypesafeApp {
 
     val workService: StacksService = StacksServiceBuilder.build(config)
 
-    val router: ItemsApi = new ItemsApi {
+    val apiConf =
```
jamieparkinson, line 30: Unless I've missed it elsewhere I think you'll need one of these https://github.com/wellcomecollection/catalogue-api/blob/fb6484df69a608c64b9555c6db8e8c71fcc8a0f3/search/src/main/scala/uk/ac/wellcome/platform/api/search/Main.scala#L29

jamieparkinson (approved):

PR #36, opened 2021-05-04
We don't need to put the context parameter on DisplayResultList 

alexwlchan:
Right now it's appearing in the Swagger documentation for the class, which is definitely wrong.  We already have a ContextResponse object for adding the context URL, so let's use it.


search/src/main/scala/uk/ac/wellcome/platform/api/search/rest/ImagesController.scala
```diff
@@ -92,13 +92,15 @@ class ImagesController(elasticsearchService: ElasticsearchService,
...
                 extractPublicUri { uri =>
                   complete(
-                    DisplayResultList(
-                      resultList = resultList,
-                      searchOptions = searchOptions,
-                      includes =
-                        params.include.getOrElse(MultipleImagesIncludes.none),
-                      requestUri = uri,
-                      contextUri = contextUri
+                    ContextResponse(
```
kenoir, line 95: 👏

kenoir (approved):

PR #38, opened 2021-05-04
Serve documents at `/` as well as `/catalogue/v2`

jamieparkinson:
Whilst preserving public URLs where they matter:
- context.json
- pagination links
- swagger
- redirects

In the course of doing it I've tried to simplify / make consistent some of the API testing logic around URL construction. Scalafmt has helpfully made some unrelated changes.

Interesting bits:
- [public URI extraction](https://github.com/wellcomecollection/catalogue-api/pull/38/files#diff-4930fb91bab87aca3f8876aca56db52dd2ca54a652c2afdfedab4dd486f14f13R27-R34)
- [routes](https://github.com/wellcomecollection/catalogue-api/pull/38/files#diff-eb7731b17eaba0803d1149beb38e133d7c11e1c13fff0a90bbb0470c3ab4d5c1R106-R113)
- [URLs used in tests](https://github.com/wellcomecollection/catalogue-api/pull/38/files#diff-e7ad1390e9a6c4ddd6aac3536fbcba9b3d21dcb43f9436a4cc894846283158f3R10-R16) - this is currently using the legacy `/catalogue/v2` root URL in the test host fixture, but has been confirmed to work at `/` and will be updated to match the behaviour when I come back round to remove the legacy routes.


search/src/main/scala/uk/ac/wellcome/platform/api/search/Router.scala
```diff
@@ -45,63 +45,70 @@ class Router(
...
+    pathPrefix("v2") {
+      concat(
+        searchRoutes,
+        pathPrefix("management") {
+          concat(
+            path("healthcheck") {
+              get {
+                complete("message" -> "ok")
+              }
+            },
```
alexwlchan, line 96: Do we need a healthcheck path in the non-legacy routes?

search/src/main/scala/uk/ac/wellcome/platform/api/search/swagger/SwaggerDocs.scala
```diff
@@ -15,28 +15,25 @@ import io.swagger.v3.oas.annotations.enums.ParameterIn
...
-
-  val version = ApiVersions.v2
-
-  val url =
-    s"${apiConfig.scheme}://${apiConfig.host}/${apiConfig.pathPrefix}/$version/"
+  val url = apiConfig.publicRootPath match {
+    case "" =>
+      s"${apiConfig.publicScheme}://${apiConfig.publicHost}/"
+    case path =>
+      s"${apiConfig.publicScheme}://${apiConfig.publicHost}$path/"
```
alexwlchan, line 29: I'm not clear why you're matching here, when dropping `${apiConfig.publicRootPath}` into the string would have the same effect.

search/src/main/scala/uk/ac/wellcome/platform/api/search/swagger/SwaggerDocs.scala
```diff
@@ -233,7 +232,9 @@ trait MultipleImagesSwagger {
             "cc-by-nc-nd",
             "cc-0",
             "pdm",
-            "copyright-not-cleared")),
+            "copyright-not-cleared"
```
alexwlchan, line 235: Let's not fix it here, as this diff is big enough, but this list is wrong – it's missing some licences, and other strings aren't actual licences. Tracked as https://github.com/wellcomecollection/platform/issues/5163

alexwlchan (approved):
I'm not going to pretend I read every line in excruciating detail, but it looks broadly correct. A couple of suggestions, and a note to me to hold merging https://github.com/wellcomecollection/catalogue-api/pull/36 until this goes in – resolving merge conflicts will be much easier on my PR.

kenoir (approved):
Looks fine to me.

search/src/main/scala/uk/ac/wellcome/platform/api/search/swagger/SwaggerDocs.scala
jamieparkinson: Oops, think this was from an earlier iteration

search/src/main/scala/uk/ac/wellcome/platform/api/search/Router.scala
jamieparkinson: No - the healthcheck is for the service itself so stays at the root. Also, more pragmatically, I don't believe we ever use it 🙃

PR #42, opened 2021-05-05
Actually add the secrets module used by Terraform

alexwlchan:
We used to put secrets in a `secrets.zip` bundle that would be unpacked by Travis, but we don't do that any more. Unfortunately, it meant I forgot to check in the `secrets` module when I added it in a recent PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,6 @@ shared_infra/.releases
 .pytest_cache
 
 # Files which could contain secrets we do not want in the repo
-secrets
-secrets.zip
 id_rsa
 id_rsa.pub
 /config
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,10 @@ id_rsa.pub
 /config
 credentials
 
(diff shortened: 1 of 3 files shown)


kenoir:
I would keep secrets.zip in the .gitignore - someone might have an old copy hanging around in their repo that could get inadvertently committed.

jamieparkinson (approved):

alexwlchan:
> I would keep secrets.zip in the .gitignore - someone might have an old copy hanging around in their repo that could get inadvertently committed.

Is that actually true? This is a brand new repo, so nothing should have been copied over – and I'm 98% sure all the Travis secrets have been burned.

I'll put it back, but with a comment explaining why it's there.

kenoir:
>Is that actually true? This is a brand new repo, so nothing should have been copied over – and I'm 98% sure all the Travis secrets have been burned.

Ah, correct. For anyone except me who filter-branched the old repo to get this one!

PR #50, opened 2021-05-05
Fetch works from Elasticsearch, not the Catalogue API

alexwlchan:
This monster of a PR is a proof-of-concept for https://github.com/wellcomecollection/catalogue-api/issues/8

It contains implementations of the items and requests API that only query for works data from Elasticsearch, and not from the Catalogue API. It’s also way too big to be reviewable – we'll need to break it up into smaller pieces, so it can be reviewed and tested properly. But it shows a way forward!

The crux is two new classes:

```scala
class WorkLookup {
  def byCanonicalId(id: CanonicalId)(index: Index): Future[Either[ElasticError, Option[Work[Indexed]]]]
}

class ItemLookup {
  def byCanonicalId(itemId: CanonicalId)(index: Index): Future[Either[ElasticError, Option[Item[IdState.Identified]]]]

  def bySourceIdentifier(itemId: SourceIdentifier)(index: Index): Future[Either[ElasticError, Option[Item[IdState.Identified]]]]
}
```

which give us the data, and then a whole bunch of plumbing to make them actually work.

Tomorrow I'll set about carving this up into smaller, more manageable pieces.


alexwlchan:
Superseded by other PRs.

PR #51, opened 2021-05-06
Add some lookup classes for Works and Items

alexwlchan:
For #8, pulled out of #50

This PR adds two new classes that we'll need to be able to query items from Elasticsearch:

```scala
class WorkLookup {
  // items API – "what are all the items on work XYZ, and what is their status?"
  def byCanonicalId(id: CanonicalId)(index: Index): Future[Either[ElasticError, Option[Work[Indexed]]]]
}

class ItemLookup {
  // requests API – "please place a hold on item XYZ"
  def byCanonicalId(itemId: CanonicalId)(index: Index): Future[Either[ElasticError, Option[Item[IdState.Identified]]]]

  // requests API – "Sierra has a hold on item ABC, what is that in catalogue land?"
  def bySourceIdentifier(itemId: SourceIdentifier)(index: Index): Future[Either[ElasticError, Option[Item[IdState.Identified]]]]
}
```

Actually plumbing these in anywhere gets quite messy, so I'll do that in separate PRs.


Makefile
```diff
@@ -12,8 +12,8 @@ PROJECT_ID = catalogue_api
 SBT_APPS = search
 SBT_NO_DOCKER_APPS = items requests
 
-SBT_DOCKER_LIBRARIES    =
-SBT_NO_DOCKER_LIBRARIES = display stacks search_common
+SBT_DOCKER_LIBRARIES    = stacks
```
jamieparkinson, line 15: What does this do / change?

common/stacks/src/main/scala/weco/api/stacks/services/ItemLookup.scala
```diff
@@ -0,0 +1,106 @@
...
+
+    elasticsearchService.findBySearch[Work[Indexed]](searchRequest).map {
+      case Left(err) => Left(err)
+      case Right(works) =>
+        Right(
+          works
+            .flatMap { _.data.items }
+            .collectFirst {
+              // The .asInstanceOf here will be a no-op at runtime, and is just so
+              // the compiler knows this will always be an Item[IdState.Identified]
```
jamieparkinson, line 48: I'm surprised the pattern match doesn't tell it this! It definitely does _sometimes_

jamieparkinson (approved):

Makefile
alexwlchan: It's because the `stacks` library has a Docker Compose file – so the tests will treat it a bit differently, and start it first.

That used to be the case, anyway…

common/stacks/src/main/scala/weco/api/stacks/services/ItemLookup.scala
alexwlchan: Yeah, it’s one of those things I expected to work. Also, I thought I'd written a much longer comment here, hmm. Will add back in the next PR.

kenoir (changes requested):
Can we discuss this before merging? I'm not sure we should have an item lookup method like this if it cannot return consistent results.

alexwlchan:
> Can we discuss this before merging? I'm not sure we should have an item lookup method like this if it cannot return consistent results.

Ah sorry, hit merge literally as you commented.

alexwlchan:
> Can we discuss this before merging? I'm not sure we should have an item lookup method like this if it cannot return consistent results.

I've reverted this on main, and created #53 to discuss some more.

PR #52, opened 2021-05-06
Add an Elasticsearch error hierarchy/tidy up type signatures

alexwlchan:
Follows #51, something I noticed while working on #50

The type signatures for methods that query Elasticsearch are… unwieldy. We hit all three of the classic monads:

```
def f(in: In): Future[Either[ElasticError, Option[T]]]
```

The `Option[T]` is there to distinguish "did we look up a thing that exists" – because elastic4s will return a "successful" GetResponse even if there isn't a document behind it.

This patch introduces our own hierarchy of error types for the Left, so we get the slightly nicer signature:

```
def f(in: In): Future[Either[ElasticsearchError, T]]
```

and that left can be carried all the way through the stack, rather than having to be handled separately at each stage.

I suspect this will be particularly nicer in the stacks-related code, where we'll get an Elasticsearch response and then do additional processing, not just return it straight to the user.


jamieparkinson (approved):
Nice change, I also do not like the current type signatures

PR #53, opened 2021-05-06
Add some lookup classes for Works and Items, take 2

alexwlchan:
I've reverted #51 in main, and reopening it here for further discussion.

### DIFF
--- a/common/stacks/src/main/scala/weco/api/stacks/services/ItemLookup.scala
+++ b/common/stacks/src/main/scala/weco/api/stacks/services/ItemLookup.scala
@@ -1,0 +1,106 @@
+package weco.api.stacks.services
+
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s.{ElasticError, Index}
+import uk.ac.wellcome.models.Implicits._
+import weco.api.search.elasticsearch.ElasticsearchService
+import weco.catalogue.internal_model.identifiers.{
+  CanonicalId,
+  IdState,
+  SourceIdentifier
+}
+import weco.catalogue.internal_model.work.{Item, Work}
+import weco.catalogue.internal_model.work.WorkState.Indexed
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class ItemLookup(elasticsearchService: ElasticsearchService)(
+  implicit ec: ExecutionContext
+) {
+
+  /** Returns an Item that corresponds to this canonical ID.
+    *
+    * Note: the output from this function may not be consistent, even against
+    * a fixed API index.  Because of merging logic, some item IDs have varying
+    * representations on different works.
+    *
+    */
+  def byCanonicalId(itemId: CanonicalId)(index: Index)
+    : Future[Either[ElasticError, Option[Item[IdState.Identified]]]] = {
+    val searchRequest =
+      search(index)
+        .query(
+          boolQuery.filter(
+            termQuery("data.items.id.canonicalId", itemId.underlying),
+            termQuery(field = "type", value = "Visible")
+          )
+        )
+        .size(1)
+
+    elasticsearchService.findBySearch[Work[Indexed]](searchRequest).map {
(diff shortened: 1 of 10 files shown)


common/stacks/src/main/scala/weco/api/stacks/services/ItemLookup.scala
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /** Returns an Item that corresponds to this source identifier.
+    *
+    * Note: the output from this function may not be consistent, even against
+    * a fixed API index.  Because of merging logic, some item IDs have varying
+    * representations on different works.
+    *
+    */
+  def bySourceIdentifier(itemId: SourceIdentifier)(index: Index)
```
kenoir, line 64: How useful is this if it cannot be consistent? If it's intended use is in the items & requesting services will an inconsistent lookup suffice?

alexwlchan: So it depends on how much information we want to expose here – we could make it consistent by tightening the output.

You want to go canonical ID ~> Sierra source identifier, so you can talk to Sierra.

You want to go Sierra source identifier ~> item canonical ID, possibly work ID (although this is a little tricky wrt Works because it's not 1:1).

If we clamped these methods to just return IDs (making them more of an IdentifierLookup), that would I think be consistent and might be better?

kenoir: > If we clamped these methods to just return IDs (making them more of an IdentifierLookup), that would I think be consistent and might be better?

Would that disrupt who you might have intended to use the work lookup?

alexwlchan: I don't *think* so – the Work lookup would be fine as-is, because it does return unambiguous results.

It's just the Items lookup that would be affected. It would mean we can't expose any other item information in the list of results; whether that's an issue depends on how much we want to return. (We don't currently put much info in the item results: https://github.com/wellcomecollection/catalogue-api/blob/main/requests/src/test/scala/uk/ac/wellcome/platform/api/requests/RequestsApiFeatureTest.scala#L121-L147

alexwlchan:
Pushed another version that relies on identifiers – hopefully this is better?

common/stacks/src/main/scala/weco/api/stacks/services/ItemLookup.scala
```diff
@@ -17,16 +17,12 @@ import scala.concurrent.{ExecutionContext, Future}
...
-  /** Returns an Item that corresponds to this canonical ID.
-    *
-    * Note: the output from this function may not be consistent, even against
-    * a fixed API index.  Because of merging logic, some item IDs have varying
-    * representations on different works.
+  /** Returns the SourceIdentifier of the item that corresponds to this
+    * canonical ID.
     *
     */
   def byCanonicalId(itemId: CanonicalId)(index: Index)
```
kenoir: Can't an item looked up this way have multiple `SourceIdentifier`s?

alexwlchan: Ah, so I meant this to be "the SourceIdentifier that corresponds to the CanonicalId" (i.e. what you'd get in the ID minter table) – but hmm, yeah, maybe it should be a list.

alexwlchan: Now I think about this – I think I do want the canonical SourceIdentifier for now; the other lists can get unwieldy.

alexwlchan: More specifically: once you start looking at multiple identifiers, you multiply the potential error states with duplicates/mismatches/unrelated identifiers. I think we could support it if we wanted to, but there's no compelling reason to look at anything but the first.

kenoir (approved):
👍

PR #57, opened 2021-05-06
Use Elasticsearch to retrieve catalogue data in the items API

alexwlchan:
For https://github.com/wellcomecollection/catalogue-api/issues/8

When the items API tries to look up the status on an item, it now fetches the associated Work from Elasticsearch, and uses that data to decide how to query the Sierra API.

Quick rundown of the major classes:

- `ItemsApi` now extends a trait `LookupItemStatus`, which takes the parameters and does the heavy lifting. This is following a pattern we have in the storage-service, where each trait is responsible for a single operation.

- The first thing `LookupItemStatus` does is try to retrieve a Work from Elasticsearch, and then it calls a method `mapVisible` from a new `SingleWorkDirectives` trait. This means we get consistent handling of works that aren't visible between the items and the search API (e.g. invisible/deleted/redirected works). You can see this at work in the items API feature test, where previously the items API would 500.

- I've brought in a variant of the SierraRecordNumber class from the pipeline repo, for a bit more consistency.

Doing Elasticsearch-based lookups in the requesting API will come in a separate patch. It'll also take some tweaking to deploy this, which again I'll do separately.


jamieparkinson (approved):
Nice

search/src/main/scala/uk/ac/wellcome/platform/api/search/rest/WorksController.scala
```diff
@@ -36,7 +36,7 @@ class WorksController(
           worksService
             .listOrSearch(index, searchOptions)
             .map {
-              case Left(err) => elasticError("Image", err)
```
kenoir, line 39: Why was this "Image" previously?

alexwlchan: Because I made a mistake, and on the search code path it doesn't actually matter.

That `documentType` variable is used to fill in the 404 message `"$documentType not found for ID $id"`, which we don't return for search responses.

common/search/src/main/scala/uk/ac/wellcome/platform/api/rest/SingleWorkDirectives.scala
```diff
@@ -0,0 +1,47 @@
...
+  implicit val ec: ExecutionContext
+
+  implicit class RouteOps(
+    work: Future[Either[ElasticsearchError, Work[WorkState.Indexed]]]) {
+    def mapVisible(
+      f: Work.Visible[WorkState.Indexed] => Future[Route]): Future[Route] =
+      work.map {
+        case Right(work: Work.Visible[Indexed]) =>
+          withFuture(f(work))
+        case Right(work: Work.Redirected[Indexed]) =>
```
kenoir, line 21: What happens when a redirected work is accessed from the items api? Is a redirection issued? Where?

kenoir (approved):
One question here to answer, but if it turns out there is a simple answer then this is good to go.

common/search/src/main/scala/uk/ac/wellcome/platform/api/rest/SingleWorkDirectives.scala
alexwlchan: It redirects you to the appropriate endpoint on the items API.

e.g. suppose A redirects to B. Then:

```
GET /works/{idA}/items
302 Found -> /works/{idB}/items

GET /works/{idB}/items
200 OK -> items on work B
```

I've added a test to show this behaviour.

PR #60, opened 2021-05-06
Query Elasticsearch when placing holds on items

alexwlchan:
For #8

This follows a similar pattern to #57 – there's a new trait `CreateRequest` that looks up an item from Elasticsearch, and delegates to the Sierra services as necessary. A few comments inline.

### DIFF
--- a/common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/source/CatalogueSource.scala
+++ b/common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/source/CatalogueSource.scala
@@ -5,14 +5,12 @@ import akka.http.scaladsl.model.Uri
 import akka.http.scaladsl.model.Uri.Path
 import uk.ac.wellcome.platform.api.common.models.Identifier
 import uk.ac.wellcome.platform.api.http.AkkaClientGet
-import weco.catalogue.internal_model.identifiers.CanonicalId
 
 import scala.concurrent.Future
 
 trait CatalogueSource {
   import CatalogueSource._
 
-  def getSearchStub(canonicalId: CanonicalId): Future[SearchStub]
   def getSearchStub(identifier: Identifier[_]): Future[SearchStub]
 }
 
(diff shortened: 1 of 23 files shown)


requests/src/main/scala/uk/ac/wellcome/platform/api/requests/RequestsApi.scala
```diff
@@ -27,22 +22,16 @@ trait RequestsApi extends CustomDirectives {
...
-            val accepted = (StatusCodes.Accepted, HttpEntity.Empty)
-            val conflict = (StatusCodes.Conflict, HttpEntity.Empty)
-
-            onComplete(result) {
-              case Success(HoldAccepted(_)) => complete(accepted)
-              case Success(HoldRejected(_)) => complete(conflict)
-              case Failure(err)             => failWith(err)
+            // TODO: We get the work ID as part of the item request, although right now
+            // it's only for future-proofing, in case it's useful later.
+            // Should we query based on the work ID?
```
alexwlchan, line 27: I think we probably should do this, but it's yet another bit of refactoring, and has some rough edges.

I'd rather get this patch merged (which replicates how we used to do item lookups), and then come back and tackle this in a separate PR.

requests/src/main/scala/weco/api/requests/responses/CreateRequest.scala
```diff
@@ -0,0 +1,47 @@
...
+
+        onComplete(result) {
+          case Success(HoldAccepted(_)) => complete(accepted)
+          case Success(HoldRejected(_)) => complete(conflict)
+          case Failure(err)             => failWith(err)
+        }
+
+      case Right(sourceIdentifier) =>
+        warn(s"Somebody tried to request non-Sierra item $itemId / $sourceIdentifier")
+        invalidRequest("You cannot request " + itemId)
```
alexwlchan: 👀 

A pattern match that only looks for items with a SierraSystemNumber identifier type? What possible future work might this extend to, I wonder aloud.

common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/source/CatalogueSource.scala
```diff
@@ -5,14 +5,12 @@ import akka.http.scaladsl.model.Uri
 import akka.http.scaladsl.model.Uri.Path
 import uk.ac.wellcome.platform.api.common.models.Identifier
 import uk.ac.wellcome.platform.api.http.AkkaClientGet
-import weco.catalogue.internal_model.identifiers.CanonicalId
```
kenoir, line 8: What's remaining before we can get rid of this entirely?

kenoir (approved):

common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/source/CatalogueSource.scala
alexwlchan: Looking up the items in the holds you already have, which I'll do in the final PR of this batch.

PR #61, opened 2021-05-06
When getting a list of holds, query from Elasticsearch 

alexwlchan:
We get a list of Sierra IDs when we ask for the list of user holds, but we need to turn that back into catalogue IDs before we return it.  This modifies the requests API to get that information from Elasticsearch, not the catalogue API.

Note: this also changes the behaviour if we can't find an ID in ES – we'll still return the hold, but without the catalogue ID.

I think we need to do more thinking about exactly what gets returned out of this API, and what it's going to be used for in the front-end. Silently dropping holds isn't great, but there may be other stuff we can do and/or need.

For https://github.com/wellcomecollection/catalogue-api/issues/8


common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/models/StacksUserHolds.scala
```diff
@@ -8,9 +10,10 @@ case class StacksUserHolds(
 )
 
 case class StacksHold(
-  itemId: ItemIdentifier[_],
```
kenoir, line 11: nice

kenoir (approved):
Very satisfying change 👍

PR #68, opened 2021-05-11
Run smoke tests from catalogue-api

kenoir:
Replaces the tests in https://github.com/wellcomecollection/integration

This change adds a buildkite pipeline to smoke test the stage deployed API.


.buildkite/pipeline.smoke.yml
```diff
@@ -0,0 +1,16 @@
+steps:
+  - label: "Smoke Tests: Catalogue API (stage)"
+    plugins:
+      - wellcomecollection/aws-assume-role#v0.2.2:
+          role: "arn:aws:iam::130871440101:role/experience-ci"
```
jamesgorrie, line 5: Is there a reason for this to be `experience-ci` over `catalogue-ci`?

smoke_tests/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "dependencies": {
+    "artillery": "1.6",
+    "artillery-plugin-expect": "^1.3.0",
+    "artillery-plugin-metrics-by-endpoint": "^0.1.1",
+    "aws-sdk": "^2.903.0",
+    "client-oauth2": "^4.3.3"
+  },
+  "scripts": {
+    "smoke_test_catalogue_api_stage": "artillery run -e stage catalogueApi.yaml",
```
jamesgorrie, line 10: nit: Js is normally camelCase/

.buildkite/pipeline.smoke.yml
```diff
@@ -0,0 +1,16 @@
...
+          run: smoke_test_catalogue_api_stage
+          command: [ "yarn", "run", "smoke_test_catalogue_api_stage" ]
+          env:
+            - AWS_ACCESS_KEY_ID
+            - AWS_SECRET_ACCESS_KEY
+            - AWS_SESSION_TOKEN
+            - AWS_REGION=eu-west-1
+
+  - label: "E2E Tests: wellcomecollection.org (prod)"
+    trigger: "experience-e2e-prod"
```
jamesgorrie, line 16: And we know that this won't be testing against the stage API?

jamesgorrie (approved):

.buildkite/pipeline.smoke.yml
kenoir: The secret is in the experience account, so in order to get it without specifying the whole ARN of the secret I'm assuming the experience role. I'm not sure how correct this is though - i'll have a think about it.

PR #72, opened 2021-05-14
Use separate accounts for search and items API access to ES

jamieparkinson:
It would be super cool to dynamically create the roles/users per environment and even with per-index permissions (rather than the `works-*` index pattern), but I'm not sure if this is possible with our elastic cloud account (?) and the existing providers https://registry.terraform.io/providers/phillbaker/elasticsearch/latest/docs/resources/xpack_role.

Anyway, this uses different accounts for each app, which have much more granular roles than the existing account:
image
image

### DIFF
--- a/terraform/modules/stack/services.tf
+++ b/terraform/modules/stack/services.tf
@@ -30,10 +30,10 @@ module "search_api" {
   }
   secrets = {
     es_host        = "elasticsearch/catalogue/private_host"
-    es_port        = "catalogue/api/es_port"
-    es_protocol    = "catalogue/api/es_protocol"
-    es_username    = "catalogue/api/es_username"
-    es_password    = "catalogue/api/es_password"
+    es_port        = "catalogue/search/es_port"
+    es_protocol    = "catalogue/search/es_protocol"
+    es_username    = "catalogue/search/es_username"
+    es_password    = "catalogue/search/es_password"
     apm_server_url = "catalogue/api/apm_server_url"
     apm_secret     = "catalogue/api/apm_secret"
   }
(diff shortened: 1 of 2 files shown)


alexwlchan:
> It would be super cool to dynamically create the roles/users per environment and even with per-index permissions

Similar to what you're describing: https://github.com/wellcomecollection/catalogue-pipeline/blob/main/pipeline/terraform/stack/elastic_pipeline_storage.tf#L32-L74

alexwlchan (approved):

jamesgorrie:
To be able to do that, I suppose you could have it as part of the cluster provisioning. I can't see a way using the [tf/ec provider](https://registry.terraform.io/providers/elastic/ec/latest/docs), but you might have a script that runs post-provisioning, [something like this](https://github.com/elastic/terraform-provider-ec/blob/9438c65eda962409493de4ae55be6953bdebd952/examples/deployment_with_init/es_config.sh#L7), that runs on apply?

Maybe something like [local-exec](https://www.terraform.io/docs/language/resources/provisioners/local-exec.html)?

jamieparkinson:
Yeah, I think that's the same as @alexwlchan's suggestion above - I'm just making these changes in service of getting the requesting app up and running so I will leave the yak unshaved for now, but it's something we should look into

alexwlchan:
Yup, although it's easier for pipeline-storage – we've just created the cluster, so Terraform has the root user/password.

You'd need to get them from elsewhere if you're adding users to an existing cluster.

You'd also want a way to delete those users when you're done – for pipeline-storage, we do it by deleting the whole cluster.

PR #83, opened 2021-05-26
Add catalogue-api cluster

kenoir:
This change is to move towards partitioning read / write load on the catalogue-api following up on a recent [incident retro](https://github.com/wellcomecollection/docs/blob/main/incident_retros/2021-05-20_ingestors.md#actions).

This PR adds a catalogue-api cluster and adds a script to provision necessary roles & users.

Future PRs will add scripts to initiate creation of follow indexes and to wire this up to the catalogue-api service. This change simply adds the cluster/secrets.


jamieparkinson:
- Will we have to (or should we) make changes to the other places that read from the catalogue cluster? ie items & requests and maybe some other things too
- Is there any way of avoiding the local-exec provisioner? eg a third party provider like [this one](https://registry.terraform.io/providers/phillbaker/elasticsearch/latest/docs/resources/xpack_user)

kenoir:
> * Will we have to (or should we) make changes to the other places that read from the catalogue cluster? ie items & requests and maybe some other things too

We will need to update all the API apps to read from the new cluster eventually.

> * Is there any way of avoiding the local-exec provisioner?

The linked provider seems only to support AWS elasticsearch - i'm more comfortable waiting for this functionality to appear in the official provider.

terraform/shared/elastic.tf
```diff
@@ -0,0 +1,133 @@
+resource "ec_deployment" "catalogue_api" {
+  name = "catalogue-api"
+
+  region                 = "eu-west-1"
+  version                = "7.13.0"
```
jamesgorrie, line 5: much fanciness!

jamesgorrie: as a point, i think this is OK - but it's worth going over the changelog as we have the most complicated use of elastic here, and could break as we're on 7.11 in prod (which we should change).

jamesgorrie (approved):

terraform/shared/elastic.tf
kenoir: yarr, we'll see when I set up a following index!

PR #88, opened 2021-05-27
Run snapshots at 11pm

kenoir:
Reduce the chance of overlap with a reindex by running snapshots out of busy hours.

This PR also adds a missing worker module left over from the api/pipeline split.


jamieparkinson (approved):

snapshots/terraform/stack/snapshot_scheduler/cloudwatch.tf
```diff
@@ -1,5 +1,6 @@
 resource "aws_cloudwatch_event_rule" "snapshot_scheduler_rule" {
-  name                = "snapshot_scheduler_rule-${var.deployment_service_env}"
-  description         = "Starts the snapshot_scheduler (${var.deployment_service_env}) lambda"
-  schedule_expression = "rate(1 day)"
+  name        = "snapshot_scheduler_rule-${var.deployment_service_env}"
+  description = "Starts the snapshot_scheduler (${var.deployment_service_env}) lambda"
+  # Between 2-3am every day
```
jamieparkinson, line 4: ?

kenoir: Oops, started of 2-3am but then I realised that might overlap with the Sierra/Calm sync (though that is less of an issue now).

alicefuzier:
I think at one point we had them run early in the morning, like 7 am or something whith the idea that they probably don't interfere with reindexes and the api in general at that time, but also if there's any problem with them someone will be available shortly to fix it.

Not saying this isn't right, just putting it out here

snapshots/terraform/stack/snapshot_scheduler/cloudwatch.tf
```diff
@@ -1,5 +1,7 @@
 resource "aws_cloudwatch_event_rule" "snapshot_scheduler_rule" {
   name                = "snapshot_scheduler_rule-${var.deployment_service_env}"
   description         = "Starts the snapshot_scheduler (${var.deployment_service_env}) lambda"
-  schedule_expression = "rate(1 day)"
+  
+  # At 11pm every day
+  schedule_expression = "cron(23 0 * * ? *)"
```
alexwlchan, line 6: Worth a comment about *why* it's at 11pm each day? You can deduce the comment just from reading the cron.

kenoir:
> but also if there's any problem with them someone will be available shortly to fix it.

🤞 shortly there will be no opportunity for them to interfere so removing the need for someone to be immediately available. I'm fairly confident this is ok for now though.

snapshots/terraform/stack/snapshot_scheduler/cloudwatch.tf
kenoir: > You can deduce the comment just from reading the cron.
Personally I find cron syntax not easy to parse.

PR #89, opened 2021-05-27
Enable logs/metrics export to logging cluster

kenoir:
This change enables export of Elasticsearch logs & metrics to the logging cluster (for the new catalogue-api cluster).

This change addresses the actions in https://github.com/wellcomecollection/docs/blob/main/incident_retros/2021-05-20_ingestors.md#actions

### DIFF
--- a/terraform/shared/elastic.tf
+++ b/terraform/shared/elastic.tf
@@ -1,3 +1,7 @@
+data "ec_deployment" "logging" {
+  id = "f4c9aca09347dad4c05f35cfbec04cdb"
+}
+
 resource "ec_deployment" "catalogue_api" {
   name = "catalogue-api"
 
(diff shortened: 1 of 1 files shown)


kenoir:
Related to: https://github.com/wellcomecollection/catalogue-pipeline/pull/1638

terraform/shared/elastic.tf
```diff
@@ -1,3 +1,7 @@
+data "ec_deployment" "logging" {
+  id = "f4c9aca09347dad4c05f35cfbec04cdb"
+}
```
alexwlchan, line 3: FYI, the logging cluster is defined in Terraform, so you could pick up this ID from that state: https://github.com/wellcomecollection/platform-infrastructure/blob/main/critical/elastic_logging_cluster.tf

alexwlchan (approved):

PR #93, opened 2021-05-28
Allow catalogue-api service secrets to vary by env

kenoir:
We will want to vary the secrets we hand to each env in order to move to a new ES cluster in stage before prod. 

This is a no-op change in terraform.

For https://github.com/wellcomecollection/platform/issues/5185

### DIFF
--- a/terraform/catalogue_api/locals.tf
+++ b/terraform/catalogue_api/locals.tf
@@ -8,4 +8,30 @@ locals {
 
   search_repository = data.terraform_remote_state.catalogue_api_shared.outputs["ecr_search_repository_url"]
   items_repository  = data.terraform_remote_state.catalogue_api_shared.outputs["ecr_items_repository_url"]
+
+  apm_secret_config = {
+    apm_server_url = "catalogue/api/apm_server_url"
+    apm_secret     = "catalogue/api/apm_secret"
+  }
+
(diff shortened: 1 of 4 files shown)


terraform/catalogue_api/locals.tf
```diff
@@ -8,4 +8,30 @@ locals {
...
+    es_port     = "catalogue/search/es_port"
+    es_protocol = "catalogue/search/es_protocol"
+    es_username = "catalogue/search/es_username"
+    es_password = "catalogue/search/es_password"
+  }
+
+  sierra_secret_config = {
+    sierra_api_key    = "stacks/prod/sierra_api_key"
+    sierra_api_secret = "stacks/prod/sierra_api_secret"
+  }
```
alexwlchan, line 36: Minor: although they're the same secret now, I would split these into two different sets, so we can assign separate Sierra API keys for the items and requests API at some future date (tracked by https://github.com/wellcomecollection/platform/issues/5130).

In particular, the requests API needs access to PII (through the patrons endpoint), whereas the items API doesn't.

kenoir: Added comment to that effect & tracking issue: https://github.com/wellcomecollection/catalogue-api/issues/95

alexwlchan (approved):

PR #98, opened 2021-06-03
Bump the internal_model for today's reindex

alexwlchan:
I've also added a script so you can bump the version without having to look it up.


jamesgorrie (approved):
Nice script

PR #108, opened 2021-06-09
Change how we handle HTTP requests to the Sierra API

alexwlchan:
This patch is a major change (and IMO improvement) to how we make HTTP requests to the Sierra API.

Previously we used a series of inherited traits:

```scala
trait AkkaGet {
  def get(…): Future[HttpResponse]
}

trait AkkaClient extends AkkaGet

trait SierraSource extends AkkaClient {
  def lookupItem() = get(…).map { resp => item}
}
```

If you want to test these interactions, you have to go through WireMock. This causes a couple of issues:

* I can't remember how to create WireMock stubs, because it's not something I do very often. I want to create a lot of these stubs to test various interactions with the holds API.
* Some of the Sierra API responses I'm mocking will include personal information (what holds a user has created) – I don't want to leak that inadvertently because I forgot to clean up my stubs.
* The interaction between a test and the associated JSON file is non-obvious. What are you actually testing?
* Every test goes through the access token handshake, even though we're not explicitly testing it in most cases.

tl;dr: I dislike WireMock and I wanted something better.

So: rather than inheritance, let's use injection. Now our interface looks more like this:

```scala
trait HttpGet {
  def get(…): Future[HttpResponse]
}

trait HttpClient extends HttpGet

trait SierraSource {
  val client: HttpClient

  def lookupItem() = client.get(…).map { resp => item}
}
```

Crucially, we can swap out this client implementation – I've got an Akka implementation that has a few WireMock tests, but the rest can be an in-memory client with an explicit list of requests and responses. I'm only partway through implementing this, but hopefully it's enough to get the idea across.

I'd like a sense check before I implement this everywhere – is this a good improvement, or is it deeply flawed?

### DIFF
--- a/common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/source/SierraSource.scala
+++ b/common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/source/SierraSource.scala
@@ -66,7 +66,7 @@ object SierraSource {
     display: String
   )
   case class SierraItemStub(
-    id: String,
+    id: SierraItemNumber,
     status: SierraItemStatusStub
   )
   case class SierraHoldRequestPostBody(
--- a/common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/source/SierraSource.scala
+++ b/common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/source/SierraSource.scala
@@ -13,6 +13,7 @@ import weco.api.stacks.http.{
   AkkaClientPost,
   AkkaClientTokenExchange
 }
+import weco.api.stacks.http.impl.AkkaHttpClient
 import weco.catalogue.source_model.sierra.identifiers.{
   SierraItemNumber,
   SierraPatronNumber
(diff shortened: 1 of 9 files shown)


alexwlchan:
I'm going to break this up into smaller pieces.

jamieparkinson:
As an aside you might find the way that the Calm API client is constructed and tested interesting for this - I did something really quite similar to what you've done here https://github.com/wellcomecollection/catalogue-pipeline/tree/main/calm_adapter/calm_api_client/src/main/scala/uk/ac/wellcome/platform/calm_api_client

alexwlchan:
> As an aside you might find the way that the Calm API client is constructed and tested interesting for this - I did something really quite similar to what you've done here https://github.com/wellcomecollection/catalogue-pipeline/tree/main/calm_adapter/calm_api_client/src/main/scala/uk/ac/wellcome/platform/calm_api_client

Mmm, and we do something similar in the storage service too. I smell an opportunity for commonisation.

kenoir (approved):
Very neat. Agreed Wiremock is not nice.

alexwlchan:
Okay, I'm going to get this passing, push it out into all the Sierra code in this repo, then yeet it up a level into scala-libs. It might also be useful for the Sierra adapter.

PR #111, opened 2021-06-10
Use new catalogue-api cluster in prod

kenoir:
Point prod search & items APIs at the new CCR follower index containing clusters.


jamesgorrie (approved):
viewed terraform diff, all good.

kenoir:
This is awaiting a review of unexpectedly poor performance in the follower index. We've a ticket open with Elastic Cloud (#00741993).

kenoir:
The performance issue was resolved by turning off an aggregations optimisation that had appeared in Elasticsearch 7.13.1. Version 7.13.2 provided a toggle to turn the optimisation off, and we have done so.

This PR is awaiting some confirmatory load tests against the stage environment.

kenoir:
Results from load tests running (2 requests per second for 120 seconds):

Screenshot 2021-06-24 at 11 23 54

Summarising these:

Screenshot 2021-06-24 at 11 24 41

I think this looks acceptable to deploy to production.

PR #113, opened 2021-06-11
Upgrade to use the newest version of libraries

alexwlchan:
Part of the refactoring for https://github.com/wellcomecollection/platform/issues/5192

I copied the classes directly out of this repo, so it was easy to swap them out.


common/stacks/src/test/scala/weco/api/stacks/services/WorkLookupTest.scala
```diff
@@ -18,38 +18,38 @@ class WorkLookupTest
...
-  it("returns Left[ElasticError] if Elasticsearch has an error") {
-    val index = createIndex
-    val future = lookup.byCanonicalId(createCanonicalId)(index)
-
-    whenReady(future) { err =>
-      err.left.value shouldBe a[IndexNotFoundError]
-      err.left.value.asInstanceOf[IndexNotFoundError].index shouldBe index.name
-    }
-  }
+//  it("returns a work with matching ID") {
```
kenoir, line 21: Should these be removed?

alexwlchan: Ah not, they should not have been commented!

kenoir (approved):

PR #114, opened 2021-06-14
Start adding scenario tests for the requesting API

alexwlchan:
The merger-style scenario tests feel like they'd work quite well here. The goal is to eventually add a scenario test for each of the outcomes in the diagram I created last week.

This patch adds the first test, plus some in-memory classes to reduce the amount of indentation and Elasticsearch fiddling we have to do in these tests.

image


requests/src/test/scala/uk/ac/wellcome/platform/api/requests/RequestingScenarioTest.scala
```diff
@@ -0,0 +1,80 @@
...
+    with IntegrationPatience {
+
+  Feature("requesting an item") {
+    Scenario("An item which is not from Sierra") {
+      Given("a physical item which is not from Sierra")
+      val item = createIdentifiedCalmItem
+
+      val lookup = new MemoryItemLookup(items = Seq(item))
+
+      withRequestsApi(lookup) { _ =>
```
jamieparkinson, line 29: nit-shed: I think it's clearer to have the whole test within the fixture block so that Given/When/Then are maximally readable

alexwlchan: I agree.

Unfortunately you need to create the item before you can invoke the fixture block, so you can't do that yet. I'm hoping to wrangle the RequestsApi so I can bin the fixture block and have them all lined up (see also: the in-memory ItemLookup saving us an Elasticsearch fixture and the `waitForPostRequest` helper), but I couldn't quite get there in this PR.

I think that'll be easier when I actually start mucking around with SierraService, which is the only fixture-y thing this is still using.

kenoir (approved):

PR #116, opened 2021-06-14
Use a single level of indentation in the scenario test

alexwlchan:
A step towards https://github.com/wellcomecollection/platform/issues/5195

This makes the scenario tests a bit nicer – they're all indented to one level now, and we don't bother starting an app that we need to tear down – instead, I'm using [route testing](https://developer.lightbend.com/guides/akka-http-quickstart-scala/testing-routes.html).

I've also added a scenario for the case when an item doesn't exist.


jamieparkinson (approved):
Nice!

PR #118, opened 2021-06-14
Fix/remove ccr following

jamesgorrie:
Just until we know what's going on with the cluster.


kenoir (approved):
I have run tests with and without CCR enabled and seen no impact, but this is ok if you want to rule it out.

jamesgorrie:
Going to leave it in as we've both seen no negative feedback

PR #119, opened 2021-06-14
Remove titles and contributors

jamesgorrie:
Reverts out the change of having `titlesAndContributors` in the index, as it's something we can add later.

This maintains the relations search as it's needed to turn off wl.org.


jamesgorrie:
Didn't perform well at al. Closing.

PR #120, opened 2021-06-15
Handle cases where items aren't available to be requested

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5195

This patch adds handling for three (admittedly unlikely) cases:

* The item has been deleted in Sierra. This could occur if somebody tries to request an item that's just been deleted in Sierra, but the deletion hasn't propagated to the catalogue API yet. We return a 400 Bad Request "This item cannot be requested".

* The item has been suppressed in Sierra. Similar to the first case – we're waiting for the update to propagate from Sierra. We return a 400 Bad Request "This item cannot be requested".

* The item doesn't exist in Sierra at all. This would be very concerning – items in Sierra only get soft-deleted, and the record gets `"deleted": true`. If the catalogue API was pointing to non-existent items, something in the pipeline is broken. We return a 500 Internal Server Error.


requests/src/test/scala/uk/ac/wellcome/platform/api/requests/RequestingScenarioTest.scala
```diff
@@ -296,6 +296,269 @@ class RequestingScenarioTest
...
+      And("the error explains why the hold is rejected")
+      withStringEntity(response.entity) {
+        assertJsonStringsAreEqual(
+          _,
+          s"""
+             |{
+             |  "@context": "$contextUrl",
+             |  "type": "Error",
+             |  "errorType": "http",
+             |  "httpStatus": 400,
```
jamieparkinson, line 477: Is there an argument that this could be a 409?

jamieparkinson (approved):

requests/src/test/scala/uk/ac/wellcome/platform/api/requests/RequestingScenarioTest.scala
alexwlchan: I think the plan is to use 409 Conflict for cases where the item is on hold to somebody else.

PR #121, opened 2021-06-15
Completely replace Wiremock with our in-memory HTTP client

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5192, follows #120


kenoir (approved):
👏 much better - less code too.

PR #122, opened 2021-06-15
add runQuery

jamesgorrie:
WIP

PR #123, opened 2021-06-15
Search profiling tool

jamieparkinson:


search_profiling/profileQuery.ts
```diff
@@ -0,0 +1,69 @@
...
+  const query = await fs.promises.readFile(
+    path.resolve(__dirname, "query-templates", `${index}.json`),
+    "utf8"
+  );
+  const aggs = await fs.promises.readFile(
+    path.resolve(__dirname, "query-templates", "aggregations.json"),
+    "utf8"
+  );
+
+  const queryHash = crypto.createHash("md5").update(query).digest("hex");
```
jamesgorrie, line 35: I've used lz-string (specifically `compressToEncodedURIComponent`) for this type of stuff before, which makes it easy to unhash

https://pieroxy.net/blog/pages/lz-string/index.html

jamieparkinson: Ooh, didn't know about that - nice

search_profiling/queryExecution.ts
```diff
@@ -0,0 +1,52 @@
...
+    password: process.env.ES_PASS!,
+  },
+});
+
+const constructQueryBody = (
+  query: string,
+  searchTerm: string,
+  aggs?: string
+): Record<string, any> => ({
+  query: JSON.parse(query.replace("{{query}}", searchTerm)),
```
jamesgorrie, line 21: `replaceAll`?

jamieparkinson: oops!

search_profiling/README.md
```diff
@@ -0,0 +1,13 @@
+# Search query profiling
+
+This provides a script which can run an elasticsearch query many times, using different search terms each time to overcome caching, and return statistics on the performance of the queries.
+
+## Usage
+
+1. Run `yarn` to install dependencies.
+2. Run `yarn refresh-templates` to get template files from the prod/stage search-templates.json endpoints.
```
kenoir, line 8: What about where we want to test custom queries not from `search-templates.json`? Might be useful to have an example custom query in here too.

kenoir: Looks like you included an example below (aggregations) - just a mention that you can do that perhaps.

kenoir (approved):
Nice.

search_profiling/README.md
jamesgorrie: We do this in rank. Ping @harrisonpim - any thoughts on if we should bring profiling into rank?

https://github.com/wellcomecollection/catalogue-api/pull/122/files#diff-3695c3f2fb64cf49101811836b159b2672287c38969db5f143e0cfadb3c39600

jamieparkinson: @kenoir this gets the files with the intention that they can be modified - I can probably improve the ergonomics of this a bit so that you can have more than one query file per index

harrisonpim: i feel like this is exactly the kind of thing we're working towards with rank, so yeah, would be great to join them up in the near future

PR #125, opened 2021-06-15
add test results to profiling

jamesgorrie:
This isn't the place for the results, but it was easier than storing it in the spreadsheet at first (will dump there now).

There are two baselines, one with the current setup on prod, the one with the current baseline on stage.

I ran a few different tests, firstly looking at the `max_expansions` on the fuzzy search. While this helped by dropping from `50 -> 40`, there seemed to be diminishing or fewer returns after that and we were still over the 500ms bar. We already have `prefix_length: 2`, which is [a recommendation when using `max_expansions`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-fuzzy-query.html).

What seems to make a significant difference is removing a piece of the query that we use to match the prefix of a title. I tried this on the prod version too and it seems to have made a difference. @harrisonpim it would be good to talk about whether we think we can rework, or drop, at least for a bit, that part of the query?

| q | stage | p95 | saving |
|---|---|---|---|
| baseline | prod | 344 | 0 |
| no prefix title | prod | 279 | 19% |
| baseline | stage | 707 | 0 |
| no prefix title | stage | 482 | 32% |
| ME 40 | stage | 600 | 16% |
| ME 30 | stage | 569 | 20% |
| ME 20 | stage | 624 | 12% |
| ME 10 | stage | 687 | 3% |


search_profiling/profileQuery.ts
```diff
@@ -30,7 +29,7 @@ const main = async () => {
   const { nSamples, index, query: queryFile, withAggs, maxPerSecond } = argv;
   const minDelay = 1000.0 / maxPerSecond;
   const query = await fs.promises.readFile(
-    path.resolve(__dirname, "query-templates", queryFile ?? `${index}.json`),
+    path.resolve(__dirname, "query-templates", queryFile || `${index}.json`),
```
jamesgorrie, line 32: Sorry about this TS was complaining.

kenoir:
I'm having difficulty following what this change is for?

jamesgorrie:
Closing, I'll move the results to the issue.

PR #126, opened 2021-06-15
remove title prefix part of works query

jamesgorrie:
Ref: https://github.com/wellcomecollection/catalogue-api/issues/127

The prefix part of the query helps us ensure that if someone types in the exact title of an work, this will definitely come up first. Shingles is another way we do this by upping the score on phrase matching.

The question is: Are we willing to drop this piece of the query to be able to release better fuzziness and collection search? In the famous words of @harrisonpim - search is a comprimise.

## Elastic profiling
https://github.com/wellcomecollection/catalogue-api/issues/127

## API services profiling
[Data source](https://docs.google.com/spreadsheets/d/1W9C9RYlLf_zrktxh0dM1EvQELMVLqTG-uEDjc3DdV3o/edit?usp=sharing)

Averaging the `p95` out we have:
| Env | p95 | %increase |
|---|---|---|
| Prod | 311 | 0 |1
| Candidate query | 685 | 65% |
| Candidate query from this PR | 413 | 25% |

While there is an increase, we have an increase of complexity, which I feel is just.

### DIFF
--- a/rank/data/queries/works-with-search-fields.ts
+++ b/rank/data/queries/works-with-search-fields.ts
@@ -28,28 +28,6 @@ export default {
         dis_max: {
           queries: [
             {
-              bool: {
-                _name: 'title prefix',
-                boost: 1000.0,
-                must: [
-                  {
-                    prefix: {
-                      'data.title.keyword': {
-                        value: '{{query}}',
-                      },
-                    },
-                  },
-                  {
-                    match_phrase: {
-                      'data.title': {
-                        query: '{{query}}',
-                      },
-                    },
-                  },
-                ],
-              },
-            },
-            {
               multi_match: {
                 _name: 'title and contributor exact spellings',
                 fields: [
--- a/rank/ranks.ts
+++ b/rank/ranks.ts
@@ -42,7 +42,7 @@ const ranks: Rank[] = [
 
       const template = {
         id: 'works-with-search-fields',
-        index: 'works-with-search-fields',
+        index: 'works-2021-06-08',
         namespace: 'works' as Namespace,
         source: { query },
       }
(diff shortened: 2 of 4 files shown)


kenoir:
Can you add a comparison between baseline and the improvement you saw in the description so that it is clear what performance improvement this change is enacting?

kenoir:
It would also be good to understand what removing the prefix part of the query will do to the search results practically. Perhaps provide some examples?

jamieparkinson:
What's the meaning of the 2/3/4 results for each query heading here?

harrisonpim:
@kenoir you should be able to see results in rank for this query compared to prod https://catalogue-api-rank-11sg28hv5-weco1.vercel.app/search?query=&rankId=works-with-search-fields

jamesgorrie:
> What's the meaning of the 2/3/4 results for each query heading here?

It was a misleading table, I've now linked to the source, and given the analysis.

harrisonpim (approved):
We definitely need some time after this to understand why performance has changed so dramatically through all of these context shifts, but i'm okay with deploying for the time being if it gives us the baseline we need for archive search

jamesgorrie:
What this part of the query was doing is:
* If a title starts with (prefix) `{q}`
* And all the terms are in the same order as `{q}`
* Match 

It was a bit of a sledge hammer which has issues (keyword matching on text fields for example). There's better ways to do this, e.g. [suggestors](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html).

An example would be searching for `The Piggle` should match `The Piggle : an account of the psychoanalytic treatment of a little girl / by D. W. Winnicott ; edited by Ishak Ramzy.` as the first result, but with this removed, it becomes the second result to `Preface to <i>The Piggle</i>` - which is something we're willing to drop for now and progress the title search further.

jamesgorrie:
@harrisonpim 

> This feels like a compromise which we don't fully understand yet

Contributing to this I feel - from a few directions - way too much work going on at once on 1 thing.

PR #139, opened 2021-06-22
create test service for rank

jamesgorrie:
* creates test service
* allows it to be accessed via HTTP and produces an API

The idea is that we run 1 test at a time for performance, but this also then opens us up to run things in CI.

e.g. 
* HTTP: `/api/test?index=works-2021-06-08&testId=precision&queryId=works&namespace=works`
* CLI: `yarn run service test index=works-2021-06-08 testId=precision queryId=works namespace=works`


harrisonpim:
Love this idea

jamesgorrie:
Screenshot 2021-06-23 at 12 58 36

@harrisonpim this still needs work, but I am loathed to move much further without review. Fancy having a peek?

harrisonpim:
Looking good! I'm finding it a bit hard to read the results though - have poked the layout a bit and come up with 👇 
Screenshot 2021-06-23 at 14 00 51

what do you think?

harrisonpim:
I feel like we could almost replace the existing `/index` and `/search` pages with this new format... I like being able to get `explain` results etc from those existing pages but it shouldn't be too hard to replicate that functionality here, right?

jamesgorrie:
> I feel like we could almost replace the existing /index and /search pages with this new format... I like being able to get explain results etc from those existing pages but it shouldn't be too hard to replicate that functionality here, right?

I don't want to do that just yet, and kind of would prefer the `index` to show the current state of play.

Completely happy with that layout!

harrisonpim:
any idea why the false positives on works are causing errors?

harrisonpim (approved):

jamesgorrie:
> any idea why the false positives on works are causing errors?

Just looking into it now

PR #142, opened 2021-06-23
CCR Cluster updates

kenoir:
Updates to the most recent version of the TF provider to get the latest version of the Elastic Cloud provider that fixes a bug where trust relationships were switched on after an update.

See https://github.com/elastic/terraform-provider-ec/blob/master/CHANGELOG.md#020-jun-15-2021

Also upgrades to the most recent version of Elastic Cloud which allows switching off a aggregation optimising feature that was causing performance issues in aggregation queries.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-7.13.2.html#bug-7.13.2

This change also adds a cluster settings update which utilises the ability to disable the troublesome optimisation.


alexwlchan (approved):

terraform/shared/scripts/create_elastic_users.py
```diff
@@ -140,6 +140,17 @@ def store_secret(secret_id, secret_value, role_arn):
             role_arn=role_arn
         )
 
+    # Configure cluster settings
+    cluster_client = elasticsearch.client.ClusterClient(es)
+
+    cluster_client.put_settings(
```
jamesgorrie, line 146: Can we use this?
https://registry.terraform.io/providers/elastic/ec/latest/docs/resources/ec_deployment#user_settings_yaml

jamesgorrie (approved):

kenoir:
This helps with: https://github.com/wellcomecollection/catalogue-api/pull/142

terraform/shared/scripts/create_elastic_users.py
kenoir: Yes we can. Good spot.

PR #145, opened 2021-06-24
add search_templates to rank

jamesgorrie:
Small PR to add a route to rank `/api/search-templates`.

This will then allow you to create local search-templates to run tests against and compare them to remote ones.


harrisonpim (approved):
✨

PR #148, opened 2021-06-24
add prototype rank interface for test service

jamesgorrie:
index2 is just a tests to see if we can generate what we want. It seems we can.

If you have run the `devInit` script you will have the local vs prod. If you add other indices, and corresponding a `{indexname}.json` - those will automagically show up.

This means we don't need a separate page for index / test.

The test are run async, which puts a little more load on the client, but it seems fine and stops the initial super long wait, but we could do with some loading.

https://user-images.githubusercontent.com/31692/123290009-d5e5e480-d508-11eb-8cd9-9e77b200e591.mov


rank/data/queries/aggregations.ts
```diff
@@ -0,0 +1,40 @@
...
+  },
+  contributors: {
+    terms: {
+      field: 'state.derivedData.contributorAgents',
+      size: 20,
+    },
+  },
+}
+
+export default aggs
```
jamesgorrie, line 40: Saving this for later.

harrisonpim: how come? what aggs stuff will we be working on soon?

harrisonpim (approved):
v cool to see the flow working 🎉 

i reckon there's still some API and UI tidying-up to be done, and some clarification of the way we use the scripts etc, but definitely moving in the right direction!

Issue #35, opened 2021-05-04
Ensure Items API display models match RFC


kenoir:
The items API should match up with what is described in https://github.com/wellcomecollection/docs/tree/master/rfcs/039-requesting-api-design.

This will include moving stacks display models into the shared display lib.

kenoir:
This needs an update to the RFC as we have decided to return a similar structure to the works API for items, only refreshing the AccessStatus to ensure data provided to the front-end is easy to parse.

PR #149, opened 2021-06-24
Add method to get AccessStatus in SierraService

kenoir:
If we intend to refresh the `AccessStatus` on items, we need a mechanism to get it.

Part of https://github.com/wellcomecollection/catalogue-api/issues/35

### DIFF
--- a/common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/SierraService.scala
+++ b/common/stacks/src/main/scala/uk/ac/wellcome/platform/api/common/services/SierraService.scala
@@ -6,10 +6,22 @@ import uk.ac.wellcome.platform.api.common.models._
 import weco.api.stacks.http.{SierraItemLookupError, SierraSource}
 import weco.api.stacks.models._
 import weco.catalogue.internal_model.identifiers.SourceIdentifier
-import weco.catalogue.internal_model.locations.{AccessCondition, AccessMethod, ItemStatus, PhysicalLocationType}
+import weco.catalogue.internal_model.locations.{
+  AccessCondition,
+  AccessMethod,
+  ItemStatus,
+  PhysicalLocationType
+}
 import weco.catalogue.source_model.sierra.SierraItemData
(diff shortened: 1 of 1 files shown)


common/stacks/src/test/scala/weco/api/stacks/services/SierraServiceTest.scala
```diff
@@ -25,6 +27,57 @@ class SierraServiceTest
...
+            identifierType = SierraSystemNumber,
+            value = "i16010176",
+            ontologyType = "Item"
+          )
+
+          val future = service.getAccessCondition(identifier)
+
+          whenReady(future) {
+            _.value shouldBe Some(
+              AccessCondition(OnlineRequest, None, None, None, None))
```
alexwlchan, line 75: You can omit the default arguments, and make this more explicit:

```suggestion
              AccessCondition(method = AccessMethod.OnlineRequest))
```

common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
```diff
@@ -33,6 +34,16 @@ class SierraService(
...
 
+  def getAccessCondition(sourceIdentifier: SourceIdentifier)
+    : Future[Either[SierraItemLookupError, Option[AccessCondition]]] = {
+    val itemNumber = SierraItemIdentifier.fromSourceIdentifier(sourceIdentifier)
+
+    sierraSource.lookupItem(itemNumber).map {
+      case Right(item) => Right(item.getAccessCondition(itemNumber))
+      case Left(err)   => Left(err)
+    }
+  }
```
alexwlchan, line 45: Do you need the pattern match, or can this be a straight map?

```suggestion
    sierraSource.lookupItem(itemNumber)
      .map { item) =>
        Right(item.getAccessCondition(itemNumber))
      }
```

alexwlchan (approved):
Couple of suggestions, but nothing substantial. ✨

common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
kenoir: Oh dear, I forgot eithers were right biased!

PR #150, opened 2021-06-24
Searchable archive descriptions

harrisonpim:
https://catalogue-api-rank-3awspvf2b-weco1.vercel.app/search?query=&rankId=works-with-archive-descriptions


rank/data/indices/works-with-archive-descriptions.ts
```diff
@@ -6,7 +6,6 @@ import { multilingualFields, shinglesFields } from './common'
 export default {
   mappings: {
     dynamic: 'strict',
-    type: "object",
```
jamesgorrie, line 9: FYI - this is here because of stupic elastic4s and how we compare JSON queries. No fuss, just letting you know.

harrisonpim: `createTestIndex` was v unhappy when this line was included - any idea whether there's a way of satisfying both?

PR #151, opened 2021-06-24
add publishing workflow to rank

jamesgorrie:
Workflow:

* `yarn devInit`
* work on query
* `yarn devPublish images images-fancy-pants-index`
* Change the query in scala till it matches the JSON
* voi la


jamesgorrie:
Next will be the update index workflow, which will be more conveluted as the index is actually stored in `catalogue-pipeline`.

rank/package.json
```diff
@@ -6,6 +6,7 @@
     "link": "vercel link",
     "lint": "next lint --fix",
     "devInit": "ts-node -r dotenv/config ./scripts/init",
+    "devPublish": "ts-node -r dotenv/config ./scripts/publish",
     "createTestIndex": "ts-node -r dotenv/config ./scripts/createTestIndex",
     "deleteTestIndex": "ts-node -r dotenv/config ./scripts/deleteTestIndex",
     "updateTestIndex": "ts-node -r dotenv/config ./scripts/updateTestIndex"
```
harrisonpim, line 12:
```suggestion
    "createIndex": "ts-node -r dotenv/config ./scripts/createIndex",
    "deleteIndex": "ts-node -r dotenv/config ./scripts/deleteIndex",
    "updateIndex": "ts-node -r dotenv/config ./scripts/updateIndex"
```
?

rank/scripts/createIndex.ts
```diff
@@ -0,0 +1,53 @@
...
+    .then((mod) => mod.default)
+    .catch(() =>
+      error(
+        `index config file does not exist. Create one at /data/indices/${indexName}.json} for better results`
+      )
+    )
+
+  info(`creating index ${indexName}`)
+  const { body: putIndexRes } = await rankClient.indices
+    .create({
```
harrisonpim, line 27: Should we reindex the prod data into these new indexes automatically? Hard to test a new mapping without any data in it...

harrisonpim (approved):

rank/scripts/createIndex.ts
jamesgorrie: I didn't as you might want to do something before you do, but we can add that tomorrow.

PR #155, opened 2021-06-28
update search interface

harrisonpim:
apologies for the nightmare PR 🦑  

- lots of shuffling of `types`, `services`, etc as I familiarise myself with recent changes
- adds [a new search page](https://catalogue-api-rank-nnbhgew4w-weco1.vercel.app/search) with a leaner interface, Hits split out as their own component, etc
- [breaks index2](https://catalogue-api-rank-nnbhgew4w-weco1.vercel.app/index2)


rank/pages/api/ccr.ts
```diff
@@ -30,8 +31,8 @@ export default async (
     return
   }
 
-  const searchTemplates = await getSearchTemplates('prod')
-  const { body: allIndices } = await rankClient.indices.get({ index: '_all' })
+  const searchTemplates = await getRemoteTemplates('prod')
+  const { body: allIndices } = await getRankClient().indices.get({ index: '_all' })
```
jamesgorrie, line 35: You can probablt have a 
```TS
const client = getRankClient()
```

rank/pages/api/search.ts
```diff
@@ -1,35 +1,32 @@
...
-  namespace?: Namespace
-}
-
-type Q = NextApiRequest['query']
-const decoder = (q: Q) => ({
-  query: q.query ? q.query.toString() : undefined,
-  rankId: q.rankId ? q.rankId.toString() : 'works-prod',
+const decoder = (req: NextApiRequest) => ({
+  query: req.query.query ? req.query.query.toString() : undefined,
+  namespace: req.query.namespace ? req.query.namespace : 'works',
```
jamesgorrie, line 16: we have `decodeNamespace` and could create a `decodeEnv`
https://github.com/wellcomecollection/catalogue-api/blob/6272d56aa00c222f9a76541ad7e8bd4e48fb2fc5/rank/types/decoder.ts#L13-L16

rank/pages/api/search.ts
```diff
@@ -38,12 +35,13 @@ export default async (
           ...template.source,
           track_total_hits: true,
           highlight: {
-            pre_tags: ['<em class="bg-yellow-200">'],
-            post_tags: ['</em>'],
+            pre_tags: ['<span class="bg-yellow-200">'],
```
jamesgorrie, line 38: Why?

rank/pages/index2.tsx
```diff
@@ -67,10 +65,10 @@ const TestRun: FC<TestRunProps> = ({ test, template }) => {
...
   const worksTemplates = searchTemplates.filter((t) => t.namespace === 'works')
   const worksTests = tests.works
-  const imagesTemplates = searchTemplates.filter(
-    (t) => t.namespace === 'images'
-  )
-  const imagesTests = tests.images
+  // const imagesTemplates = searchTemplates.filter(
+  //   (t) => t.namespace === 'images'
+  // )
+  // const imagesTests = tests.images
```
jamesgorrie, line 71: Any reason for this?

rank/services/test.ts
```diff
@@ -29,26 +25,48 @@ type Props = {
 async function service({
   namespace,
   testId,
+  env,
```
jamesgorrie, line 28: What I was picturing is evatually having all this information stored in the search template so you don't need to have all these floating values everywhere. This is OK for now though.

jamesgorrie (approved):
Much tidier - really nice search too.

rank/pages/api/search.ts
harrisonpim: 👍

rank/pages/api/search.ts
harrisonpim: Purely because it's a bit easier on my eyes... I found myself squinting and rereading the highlights multiple times, when they were also italicised, but the highlight by itself seemed fine. 

Happy to revert if you want though?

rank/services/test.ts
harrisonpim: yeah, i agree tbh - think we're both on the same page about where all this data should be going eventually

rank/pages/api/search.ts
harrisonpim: something like this?
```ts
const decoder = (req: NextApiRequest) => ({
  query: req.query.query ? decodeString(req.query, 'query') : undefined,
  namespace: req.query.namespace ? decodeNamespace(req.query.namespace) : 'works',
  env: req.query.env ? decodeEnv(req.query.env) : 'prod',
})
```

rank/pages/index2.tsx
harrisonpim: Nope, just making a mess 😬

rank/pages/api/search.ts
jamesgorrie: Yea, I was working on a wrapper method, but couldn't quite get the typing correct - similar to:

```ts
{ query: optional(decodeString(req.query)) }
```

harrisonpim: ah nice - will keep this in mind

PR #161, opened 2021-06-28
start adding CLI tests to rank

jamesgorrie:
Having the results in the browser is useful, and can give a tonne of extra information.

This allows us however to just runs test in the CLI, thus get them into CI too.

Screenshot 2021-06-29 at 09 36 02


rank/data/tests/images.ts
```diff
@@ -117,26 +117,31 @@ const tests: Test[] = [
         query: 'monsters',
         ratings: ['n7r5s65w'],
         description: "shouldn't match 'Monastery'",
+        knownFailure: true,
```
harrisonpim, line 120: The idea of labelling tests as 'known failures' feels funky to me, even if it is the way that we've operated IRL. Conscious that these labels might rot if we're not paying close attention.

We've also talked in the past about using an overall threshold score to let the system pass as a whole. I can see that adding known failures is simpler to implement than a set of thresholds, but i'm not sure what will be best long term. What's your feeling about this stuff at the moment?

harrisonpim (approved):

rank/data/tests/images.ts
jamesgorrie: Yea, it felt odd. Let's see how we feel about uncommenting it when we add the new collections stuff.

PR #184, opened 2021-07-08
Bump version in step with pipeline 

jamesgorrie:
After conversation with @alicefuzier and @kenoir we realised that the version bump should be whatever the index says it has been created with, rather than latest. 

The latest version is pulled from `catalogue-api.ingested-works-{date}._meta`.

This change does that.


project/Dependencies.scala
```diff
@@ -13,7 +13,7 @@ object WellcomeDependencies {
     val monitoring = defaultVersion
     val storage = defaultVersion
     val elasticsearch = defaultVersion
-    val internalModel = "5021.6fbf97350117e797abb25ab41f1a00d20d7899e5"
+    val internalModel = "4988.884bf55a45d76171db88112142b2ee8dd6e37f46"
```
kenoir, line 16: `main` now contains code that depends on the most recent version, so this build will fail.

kenoir, line 16:
```suggestion
    val internalModel = "5021.6fbf97350117e797abb25ab41f1a00d20d7899e5"
```

kenoir: This will allow us to merge this PR.

kenoir (approved):
Nice script, but the internal model needs rolling forward for the build to pass.

PR #189, opened 2021-07-14
add TEI tests to rank

harrisonpim:
[you can see the new tests running on the new indexes here](https://catalogue-api-rank-fkducopbi-weco1.vercel.app/index2?worksIndex=works-indexed-2021-07-13&imagesIndex=images-indexed-2021-07-13)

### DIFF
--- a/rank/docs/arabic-examples.md
+++ b/rank/docs/arabic-examples.md
@@ -1,8 +1,0 @@
-# Arabic examples
-
-These examples should be added to rank when the TEI work has been merged, to make sure we're analysing arabic text correctly.
-
-| search term | matched term in document | work                                                                                                          |
-| ----------- | ------------------------ | ------------------------------------------------------------------------------------------------------------- |
-| لكشف        | الكشف                    | https://github.com/wellcomecollection/wellcome-collection-tei/blob/master/Arabic/WMS_Arabic_256.xml           |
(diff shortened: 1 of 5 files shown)


rank/docs/arabic-examples.md
```diff
@@ -1,8 +0,0 @@
-# Arabic examples
-
-These examples should be added to rank when the TEI work has been merged, to make sure we're analysing arabic text correctly.
-
-| search term | matched term in document | work                                                                                                          |
-| ----------- | ------------------------ | ------------------------------------------------------------------------------------------------------------- |
-| لكشف        | الكشف                    | https://github.com/wellcomecollection/wellcome-collection-tei/blob/master/Arabic/WMS_Arabic_256.xml           |
```
alexwlchan, line 7: Is it worth putting these links somewhere in the test config? They seem like useful context.

harrisonpim: Maybe... 

We already include the work ID and a short description of the match we're trying to make, and we haven't needed quick access source data in the course of the work so far. We usually think of the target works as unchangeable from rank's POV, but knowledge of source works is worth keeping in mind with the ongoing questions about matching/merging. 

I'm going to leave it out of this PR but might come back to add source works more comprehensively across our test cases another time.

harrisonpim: pronge @jamesgorrie for any thoughts on ☝️

jamesgorrie: What are the links for? As far as I can see it's an example, but now the work is in the API, we'll have the link from the ratings which is the work ID?

jamesgorrie (approved):

PR #191, opened 2021-07-14
Get multiple items from the sierra API

kenoir:
Reduce the possible number of requests going to the Sierra API for situations where works have many items.

This change updates only the `SierraSource` class. A future PR will use the new functionality in the Items API.

For https://github.com/wellcomecollection/platform/issues/5165


common/stacks/src/main/scala/weco/api/stacks/http/SierraSource.scala
```diff
@@ -31,33 +39,82 @@ class SierraSource(client: HttpClient with HttpGet with HttpPost)(
...
+  def unmarshalOKResponse[T](
+    response: model.HttpResponse
+  )(implicit um: Unmarshaller[model.HttpResponse, T]): Unmarshalled[T] =
+    Unmarshal(response)
+      .to[T]
+      .map(Right(_))
+      .recover {
+        case t: Throwable =>
+          Left(SierraItemLookupError.ItemHasNoStatus(t))
+      }
```
alexwlchan, line 61: I wonder if a pattern match would be neater than a series of `map … recover` here?

alexwlchan (commented):
Looks good so far. 👍

common/stacks/src/main/scala/weco/api/stacks/http/SierraSource.scala
```diff
@@ -31,33 +37,67 @@ class SierraSource(client: HttpClient with HttpGet with HttpPost)(
...
+    CirceMarshalling.fromDecoder[SierraItemDataEntries]
+
   private implicit val umErrorCode: Unmarshaller[HttpEntity, SierraErrorCode] =
     CirceMarshalling.fromDecoder[SierraErrorCode]
 
+  /** Returns data for a list of items
+    *
+    * Note: if some of the IDs requested do not exist, they will
+    * not appear in the results list. It's up to the consumer to
+    * identify that and react to it as required.
```
alexwlchan, line 51: Should we drop a warning log for this? It's not something that should happen very often in practice.

common/stacks/src/main/scala/weco/api/stacks/http/SierraSource.scala
```diff
@@ -31,33 +37,67 @@ class SierraSource(client: HttpClient with HttpGet with HttpPost)(
...
+  }
+
+  /** Returns data for a single item
+    */
   def lookupItem(
     item: SierraItemNumber
   ): Future[Either[SierraItemLookupError, SierraItemData]] =
     for {
-      resp <- client.get(
+      response <- client.get(
```
alexwlchan, line 89: I think you could define this function in terms of your new function:

```scala
for {
  lookupResult <- lookupItemEntries(items = Seq(item))

  item = lookupResult.entries match {
    case Seq(it) => Right(it)
    case Nil => Left(SierraItemLookupError.ItemNotFound)
    case _ => Left(SierraItemLookupError.UnknownError(new Throwable(s"Multiple items found for ID $item")))
  }
} yield item
```

common/stacks/src/main/scala/weco/api/stacks/http/SierraSource.scala
```diff
@@ -31,33 +37,67 @@ class SierraSource(client: HttpClient with HttpGet with HttpPost)(
...
+
+    val idList = items
+      .map(_.withoutCheckDigit)
+      .mkString(",")
+
+    for {
+      response <- client.get(
+        path = Path("v5/items"),
+        params = Map(
+          "id" -> idList,
```
alexwlchan, line 65: Is there a limit to how many items we can fetch in a single response? There's a work with 500+ items.

common/stacks/src/main/scala/weco/api/stacks/http/SierraSource.scala
kenoir: I've reconsidered this and now return an error state reporting on what's missing and what's returned. Realising Sierra would 404 in the case of no matches made me think we should provide a slightly nicer interface, rather than pushing that weirdness up to consumers.

common/stacks/src/main/scala/weco/api/stacks/http/SierraSource.scala
kenoir: Probably. The API documentation doesn't specify a limit AFAICT but I bet there is one. I'm not going to account for that in this PR but will test and update in the future if required.

common/stacks/src/main/scala/weco/api/stacks/http/SierraSource.scala
```diff
@@ -18,50 +18,104 @@ import weco.catalogue.source_model.sierra.identifiers.{
...
+    item: SierraItemNumber
+  ): Future[Either[SierraItemLookupError, SierraItemData]] = {
+    lookupItemEntries(Seq(item)).map {
+      case Right(itemDataEntries) =>
+        Right(itemDataEntries.entries.head)
+      case Left(SierraItemLookupError.MissingItems(_, _)) =>
+        Left(SierraItemLookupError.ItemNotFound)
+      case Left(value) => Left(value)
+    }
+  }
```
alexwlchan, line 118: nit: lose the parens:

```suggestion
  ): Future[Either[SierraItemLookupError, SierraItemData]] =
    lookupItemEntries(Seq(item)).map {
      case Right(itemDataEntries) =>
        Right(itemDataEntries.entries.head)
      case Left(SierraItemLookupError.MissingItems(_, _)) =>
        Left(SierraItemLookupError.ItemNotFound)
      case Left(value) => Left(value)
    }
```

alexwlchan (approved):

PR #193, opened 2021-07-14
add tabulated interface to rank

jamesgorrie:
Just finishing off a PR from some discussion @harrisonpim and I had re: rank interface.
Needs a few bits and bobs, but we've halted dev on this for now.

rank_tabulated


rank/components/H.tsx
```diff
@@ -2,8 +2,18 @@ import { FC } from 'react'
...
   return (
-    <h2 className="text-xl mt-1 mb-2 border-t border-gray-200">{children}</h2>
+    <h2 className="text-3xl mt-1 mb-4 border-l-4 pl-2 border-blue-300">
+      {children}
+    </h2>
   )
 }
 
-export { H2 }
+const H3: FC = ({ children }) => {
```
harrisonpim, line 11: Out of interest, what's the benefit for creating a set of `H` components here instead of just setting these styles globally in `/styles`, eg:

```
h2 {
  @apply text-3xl mt-1 mb-4 border-l-4 pl-2 border-blue-300;
}

h3 {
  @apply text-lg mt-1 mb-2 border-l-4 pl-2 border-pink-300;
}
```

afaik we don't use any headings which aren't intended to have these styles applied to them

rank/services/search-templates.ts
```diff
@@ -57,7 +61,7 @@ export async function getRemoteTemplates(env: Env): Promise<SearchTemplate[]> {
     remoteTemplates = json.templates.map((template) => {
       const namespace = getNamespaceFromIndexName(template.index)
       return {
-        id: `${env}/${namespace}/${template.index}`,
+        id: `${namespace}/${env}/${template.index}`,
```
harrisonpim, line 64: I feel like the `namespace` value might be redundant here, because we'll never want to run `images` tests against a `works` index, or vice-versa... 

If we know that the index name is `works-2021-07-13`, could we just get the namespace from that? 

Can we safely assume that the index name will always start with `works` or `images`?

rank/components/SearchTemplateTests.tsx
```diff
@@ -0,0 +1,119 @@
...
+                <td className="border border-pink-200 px-4 py-2">{query}</td>
+                {result.map((result) => (
+                  <td
+                    // This key is definitely wrong, annoyingly I can't find out how to construct something unique here.
+                    key={`${Math.random().toString(36).substring(7)}`}
+                    className={`w-6 border border-pink-200 bg-${resultColor(
+                      result.result.pass
+                    )}`}
+                  >
+                    <Result pass={result.result.pass} />
```
harrisonpim, line 78: I think the only other thing I'd want here is a link from the test to the equivalent `/search` query, so that the results themselves can be quickly inspected, but this is already a huge improvement and it should be easy enough to reconstruct the queries on the `/search` page as-is.

harrisonpim (approved):

rank/components/H.tsx
jamesgorrie: Basically so that you can use them if you want, but not if you don't want. Setting global styles always makes this incredible faffy, and lands up with loads of overrides. 

We could create a style `.h2` sa above - and then apply that to anything, which maybe fits the [utility first](https://tailwindcss.com/docs/utility-first) mentality, but I've gone for the [extracting components](https://tailwindcss.com/docs/extracting-components) method.

rank/services/search-templates.ts
jamesgorrie: I didn't want to make that assumption and it makes the code more complex. I'd prefer the split in logic.

We have a set of tests, the use a query from remote or local that is run on an index. `namespace/env/index`.

Creating any other rules has previously landed up with us building in more logic than is necessary. 

The tests will just fail if you get the wrong combo.

PR #197, opened 2021-07-20
Use lookupItemEntries in the Items API

kenoir:
Reduce the possible number of requests going to the Sierra API for situations where works have many items.

This change updates the Items API.

For https://github.com/wellcomecollection/platform/issues/5165

Follows: https://github.com/wellcomecollection/catalogue-api/pull/191

### DIFF
--- a/common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
+++ b/common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
@@ -36,7 +36,7 @@ class SierraService(
 
   def getAccessConditions(
                            itemNumbers: Seq[SierraItemNumber]
-                         ): Future[Either[SierraItemLookupError, Map[SierraItemNumber, Option[AccessCondition]]]] = {
+                         ): Future[Either[SierraItemLookupError, Map[SierraItemNumber, AccessCondition]]] = {
     for {
       itemEither <- sierraSource.lookupItemEntries(itemNumbers)
       accessCondition = itemEither.map { _.entries.map(
@@ -194,7 +194,7 @@ class SierraService(
     }
 
   implicit class ItemDataOps(itemData: SierraItemData) {
-    def getAccessCondition: Option[AccessCondition] = {
+    def getAccessCondition: AccessCondition = {
 
       val location: Option[PhysicalLocationType] =
         itemData.fixedFields
@@ -215,11 +215,8 @@ class SierraService(
     }
 
     def allowsOnlineRequesting: Boolean = {
-      getAccessCondition match {
-        case Some(ac) if ac.method == AccessMethod.OnlineRequest =>
-          true
-        case _ => false
-      }
+      val accessCondition = getAccessCondition
+      accessCondition.method == AccessMethod.OnlineRequest
     }
   }
 
--- a/common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
+++ b/common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
@@ -34,15 +34,15 @@ class SierraService(
 )(implicit ec: ExecutionContext)
     extends Logging {
 
-  def getAccessCondition(
-    sourceIdentifier: SourceIdentifier
-  ): Future[Either[SierraItemLookupError, Option[AccessCondition]]] = {
-    val itemNumber = SierraItemIdentifier.fromSourceIdentifier(sourceIdentifier)
+  def getAccessConditions(
+    sourceIdentifiers: Seq[SourceIdentifier]
+  ): Future[Either[SierraItemLookupError, Map[SierraItemNumber, Option[AccessCondition]]]] = {
+    val itemNumbers =sourceIdentifiers.map(SierraItemIdentifier.fromSourceIdentifier(_))
 
     for {
-      itemEither <- sierraSource.lookupItem(itemNumber)
-      accessCondition = itemEither.map { item =>
-        item.getAccessCondition(itemNumber)
+      itemEither <- sierraSource.lookupItemEntries(itemNumbers)
+      accessCondition = itemEither.map { itemDataEntries =>
+        itemDataEntries.entries.map(item => item.id -> item.getAccessCondition()).toMap
       }
     } yield accessCondition
   }
@@ -195,13 +195,13 @@ class SierraService(
     }
 
   implicit class ItemDataOps(itemData: SierraItemData) {
-    def getAccessCondition(id: SierraItemNumber): Option[AccessCondition] = {
+    def getAccessCondition(): Option[AccessCondition] = {
 
       val location: Option[PhysicalLocationType] =
         itemData.fixedFields
           .get("79")
           .flatMap(_.display)
-          .flatMap(name => SierraPhysicalLocationType.fromName(id, name))
+          .flatMap(name => SierraPhysicalLocationType.fromName(itemData.id, name))
 
       // The bib ID is used for debugging purposes; the bib status is only used
       // for consistency checking. We can use placeholder data here.
--- a/common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
+++ b/common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
@@ -2,24 +2,13 @@ package weco.api.stacks.services
 
 import akka.stream.Materializer
 import grizzled.slf4j.Logging
-import weco.api.stacks.http.{SierraItemLookupError, SierraSource}
+import weco.api.stacks.http.{SierraItemDataEntries, SierraItemLookupError, SierraSource}
 import weco.api.stacks.models._
 import weco.catalogue.internal_model.identifiers.SourceIdentifier
-import weco.catalogue.internal_model.locations.{
-  AccessCondition,
-  AccessMethod,
-  PhysicalLocationType
-}
+import weco.catalogue.internal_model.locations.{AccessCondition, AccessMethod, PhysicalLocationType}
 import weco.catalogue.source_model.sierra.SierraItemData
-import weco.catalogue.source_model.sierra.identifiers.{
-  SierraBibNumber,
-  SierraItemNumber,
-  SierraPatronNumber
-}
-import weco.catalogue.source_model.sierra.rules.{
-  SierraItemAccess,
-  SierraPhysicalLocationType
-}
+import weco.catalogue.source_model.sierra.identifiers.{SierraBibNumber, SierraItemNumber, SierraPatronNumber}
+import weco.catalogue.source_model.sierra.rules.{SierraItemAccess, SierraPhysicalLocationType}
 import weco.http.client.{HttpClient, HttpGet, HttpPost}
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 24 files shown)


items/src/main/scala/weco/api/items/services/ItemUpdateService.scala
```diff
@@ -23,40 +26,57 @@ class ItemUpdateService(
...
+      .getAccessConditions(sierraItemSourceIdentifiers.keys.toSeq)
+      .map {
+        case Right(accessConditions) =>
+          updateAccessConditions(sierraItemSourceIdentifiers, accessConditions)
+        case Left(err) =>
+          error(msg = f"Couldn't refresh items on: ${work.id} got error $err")
+          sierraItems
+      }
+
+    updatedSierraItems.map(_ ++ otherItems)
```
alexwlchan, line 80: You linked to this function as particularly knotty. Perhaps it would help to think of Sierra as 1 of multiple sources we might implement requesting for, rather than a special case?

I was about to suggest some sort of match statement that goes like this:

```scala
items.map {
  case sierraItem => updateSierraItem(sierraItem)
  case calmItem => updateCalmItem(calmItem)
  case item => Future.successful(item)
}
```

but of course, the whole point of this work is to update items in bulk!

How about something like this:

```scala
case class Item(id: String, scheme: String)

val items = Seq(
  Item(id = "b123", scheme = "Sierra"),
  Item(id = "b124", scheme = "Sierra"),
  Item(id = "A/B/C", scheme = "Calm"),
  Item(id = "b125", scheme = "Sierra"),
  Item(id = "V0000001", scheme = "Miro")
)

val itemsByScheme = items.groupBy { _.scheme }
// HashMap(
//   Miro -> List(Item(V0000001,Miro)),
//   Calm -> List(Item(A/B/C,Calm)),
//   Sierra -> List(Item(b123,Sierra), Item(b124,Sierra), Item(b125,Sierra))
// )

itemsByScheme.map {
  case (Sierra, sierraItems) => updateSierraItems(sierraItems)
  case (Calm, calmItems) => updateCalmItems(calmItems)
  case (_, items) => items
}
```

I might also throw in a `zipWithIndex` somewhere to preserve the original item ordering, which currently is being lost.

common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
```diff
@@ -34,15 +34,18 @@ class SierraService(
...
     for {
-      itemEither <- sierraSource.lookupItem(itemNumber)
-      accessCondition = itemEither.map { item =>
-        item.getAccessCondition(itemNumber)
+      itemEither <- sierraSource.lookupItemEntries(itemNumbers)
+      accessCondition = itemEither.map {
+        _.entries
+          .map(
+            item => item.id -> item.getAccessCondition
+        ) toMap
```
alexwlchan, line 48: Now looking at this function, hmm.

You're preserving the `SierraItemLookupError` here, and I'm not sure that's correct – if we know about some of the items, is it better to do a partial update or hard fail? I think a partial update is better, but that's difficult to communicate – without propagating the error, how do we distinguish between "we couldn't get updated item data" and "we got the updated item data, and there's now no access condition".

Which made me wonder… is there ever a case in which we won't return an access condition?

Spoiler: no. https://github.com/wellcomecollection/catalogue-pipeline/pull/1778

If you pull in that patch, you instantly lose a layer of `Option[…]` here – and I think it enables you to tidy up this signature further, and do something like:

```scala
def getUpdatedAccessConditions(
  itemNumbers: Seq[SierraItemNumber]
): Future[Map[SierraItemNumber, AccessCondition]]
```

Inside this function you log a warning about any missing items, but you don't let the error propagate any further (at least, not directly).

You return a Map of access conditions that should replace the conditions on the existing items. If an item number is missing from this Map, then the caller should keep whatever condition was already there from the Catalogue API.

common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
```diff
@@ -34,17 +38,25 @@ class SierraService(
...
     extends Logging {
 
-  def getAccessCondition(
-    sourceIdentifier: SourceIdentifier
-  ): Future[Either[SierraItemLookupError, Option[AccessCondition]]] = {
-    val itemNumber = SierraItemIdentifier.fromSourceIdentifier(sourceIdentifier)
-
+  def getAccessConditions(
+    itemNumbers: Seq[SierraItemNumber]
+  ): Future[Map[SierraItemNumber, AccessCondition]] = {
```
alexwlchan, line 43: This is much simpler now! :D

common/stacks/src/test/scala/weco/api/stacks/services/SierraServiceTest.scala
```diff
@@ -77,12 +77,135 @@ class SierraServiceTest
...
+            }
+          )
+
+          val itemNumber = SierraItemIdentifier.fromSourceIdentifier(
+            SourceIdentifier(
+              identifierType = SierraSystemNumber,
+              value = "i16010176",
+              ontologyType = "Item"
+            )
+          )
```
alexwlchan, line 142:
```suggestion
          val itemNumber = SierraItemNumber("1601017")
```

Isn't that equivalent but with less round tripping?

common/stacks/src/test/scala/weco/api/stacks/services/SierraServiceTest.scala
```diff
@@ -77,12 +77,135 @@ class SierraServiceTest
...
+
+          val future = service.getAccessConditions(Seq(itemNumber))
+
+          whenReady(future) {
+            _ shouldBe Map.empty
```
alexwlchan, line 208: Good set of test cases. 👍

items/src/main/scala/weco/api/items/Main.scala
```diff
@@ -32,7 +32,13 @@ object Main extends WellcomeTypesafeApp {
 
     val elasticClient = ElasticBuilder.buildElasticClient(config)
     val sierraService = SierraServiceBuilder.build(config)
-    val itemUpdateService = new ItemUpdateService(sierraService)
+
+    // To add a new item updater, implement ItemUpdater and add it to the list here
```
alexwlchan, line 36:
```suggestion
    // To add an item updater for a new service, implement ItemUpdater and add it to the list here
```

items/src/main/scala/weco/api/items/services/ItemUpdateService.scala
```diff
@@ -1,62 +1,91 @@
...
 import weco.catalogue.internal_model.work.{Item, Work, WorkState}
 
 import scala.concurrent.{ExecutionContext, Future}
 
+trait ItemUpdater {
+  val identifierType: IdentifierType
+  def updateItems(
+    items: Seq[Item[IdState.Identified]]
+  ): Future[Seq[Item[IdState.Identified]]]
+}
```
alexwlchan, line 17: Minor: move this to a new file?

Nice idea to put the `identifierType` on the ItemUpdater trait.

items/src/main/scala/weco/api/items/services/ItemUpdateService.scala
```diff
@@ -1,62 +1,91 @@
...
+    val items = itemsMap.values.toSeq
+
+    f(items).flatMap { updatedItems =>
+      val updatedItemsMap = buildItemsMap(updatedItems)
+      val updatedItemsWithIndex = indexLookup.map {
+        case (srcId, index) =>
+          updatedItemsMap.get(srcId).map { updatedItem =>
+            (updatedItem, index)
+          }
+      } flatten
```
alexwlchan, line 52: I think I know what's going on here and it's correct, but it's still a bit gnarly. Possibly I'm tired, but a few comments or type signatures would help me read this.

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -0,0 +1,64 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+class SierraItemUpdater(sierraService: SierraService)(
+  implicit executionContext: ExecutionContext
+) extends ItemUpdater
+    with Logging {
+
+  val identifierType = IdentifierType.SierraSystemNumber
+
+  private def updateLocations(
```
alexwlchan, line 23: Minor: I'd call this updateAccessCondition, maybe, which describes what it does – updateLocations is more of an implementation detail.

alexwlchan (commented):
My brain is too melty to review properly today, but here's what I've got.

common/stacks/src/test/scala/weco/api/stacks/services/SierraServiceTest.scala
kenoir: Good spot, in this case i've just mindlessly wrapped the code with `SierraItemIdentifier.fromSourceIdentifier` when I needed a `SierraItemNumber`

items/src/main/scala/weco/api/items/services/ItemUpdateService.scala
kenoir: Completely fair. I think this function and `updateItems` could do with some comments. There might be a little more work to make the code easier to read, but what's going on is compact and complicated, warranting a bit of written explanation.

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -0,0 +1,84 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/** Updates the AccessCondition of sierra items
+  *
+  *  This provides an up to date view on whether a hold
+  *  can be placed on an item.
+  *
+  *  @param sierraService
+  *  @param executionContext
```
alexwlchan, line 22: Do we need these @​param declarations? Are they adding anything?

items/src/test/scala/weco/api/items/fixtures/ItemsApiGenerators.scala
```diff
@@ -30,43 +34,63 @@ trait ItemsApiGenerators
...
+    testWith(new ItemUpdateService(itemUpdaters))
+
+  def buildEntry(
+    sierraItemNumber: SierraItemNumber,
+    deleted: String = "false",
+    suppressed: String = "false",
+    holdCount: Int = 0
+  ) = f"""
+                        |{
+                        |  "id": "${sierraItemNumber.withCheckDigit}",
```
alexwlchan, line 58:
```suggestion
                        |  "id": "${sierraItemNumber.withoutCheckDigit}",
```

Item IDs from Sierra don't include the check digit. I know this is copy-pasted from previous code, but the previous code was wrong.

items/src/test/scala/weco/api/items/services/ItemUpdateServiceTest.scala
```diff
@@ -85,59 +90,137 @@ class ItemUpdateServiceTest
...
+        items shouldBe orderedItems
+      }
+
+      whenReady(itemUpdateService.updateItems(workWithItemsBackward)) { items =>
+        items shouldBe reversedItems
+      }
+    }
+  }
+
+  it("detects if the item updater returns inconsistent results") {
```
alexwlchan, line 138: What does "inconsistent results" mean? Is this incomplete results (not all items were returned), or something else?

items/src/main/scala/weco/api/items/services/ItemUpdateService.scala
```diff
@@ -1,62 +1,110 @@
...
+                  itemsWithIndex = itemsWithIndex,
+                  updateFunction = updater.updateItems
+                )
+            )
+            .getOrElse(Future(itemsWithIndex))
+        case (None, itemsWithIndex) =>
+          Future(itemsWithIndex)
+      }
+      // unzipWithIndex
+    } map (_.flatten.toList.sortBy(_._2).map(_._1))
```
alexwlchan, line 109: Would be nice to use named variables rather than `_2` and `_1` here.

alexwlchan (approved):
Couple of comments, but mostly looks good to me.

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
kenoir: It might make it look nice if anyone ever generated scaladocs (the template is auto generated in Intellij), but it's unlikely. Will remove.

items/src/test/scala/weco/api/items/services/ItemUpdateServiceTest.scala
kenoir: Too few or the item updater has returned items with differing IDs. Inconsistent meaning the items updated can't be paired with the items handed for update.

I'll think about how to better convey that.

items/src/main/scala/weco/api/items/services/ItemUpdateService.scala
kenoir: tuples with two things should have better names for accessing their elements :( I wonder if there is a way to do this without wrapping in a case class.

kenoir: The answer is: deconstruct the tuples and name them:

```scala
case (thingOne, _) => thingOne
```

PR #200, opened 2021-07-23
Add a test for a user who is barred

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4295

I wrote out a long list of test cases in February 2020, and we forgot to implement one of them.

### DIFF
--- a/requests/src/test/scala/weco/api/requests/RequestingScenarioTest.scala
+++ b/requests/src/test/scala/weco/api/requests/RequestingScenarioTest.scala
@@ -921,6 +921,97 @@ class RequestingScenarioTest
         )
       }
     }
+
+    Scenario("A user whose account is barred") {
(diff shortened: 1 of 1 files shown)


kenoir (approved):
Very timely. @pollecuttn was asking me what the behaviour of this was.

alexwlchan:
> @pollecuttn was asking me what the behaviour of this was.

So unfortunately we can't really tell from Sierra directly – we have to guess. In particular this error:

```
{
  "code": 132,
  "specificCode": 2,
  "httpStatus": 500,
  "name": "XCirc error",
  "description": "XCirc error : There is a problem with your library record.  Please see a librarian."
}
```

is the same whether your record has been manually blocked, or because you're at your hold limit. We return it as a 500 error because we can't actually be sure what's up here.

PR #202, opened 2021-07-26
Update internal model

kenoir:
This PR updates the internal model, and makes changes required from that update.

Extracted from https://github.com/wellcomecollection/catalogue-api/pull/197

- Add implicit for language code generation:
  - Fix for an update in the internal model breaking ScalaCheck generation of a work. Language codes got additional restrictions not reflected by the available implicits in scope in the `DisplayWorkTest`.
  - This change also adds some more comments to make it clearer what is happening with Work generation.

- Update to replace `Option[AccessCondition]` with `AccessCondition`


alexwlchan (approved):
I wonder if it would be a good idea to upstream this to the pipeline repo, then re-import it here?

PR #205, opened 2021-07-27
add tooling for internal model

jamesgorrie:
Adds tooling for checking the internal model and it's compatibility with the index name. This should catch it before it goes out and is deployed, which is a massive time suck.

I played around with bumping the `internalModel` automatiocally, but it seemed like a silly idea.

If we're using the `bump` tool, this shouldn't be needed, but we have got caught before...


.buildkite/pipeline.yml
```diff
@@ -1,6 +1,18 @@
 steps:
   - command: .buildkite/scripts/run_autoformat.py
     label: "autoformat"
+  - label: "Internal model tool: check compatibility"
```
kenoir, line 4: You might want to run this on branches as well as in `main` (if that's the intention - i'm guessing it is from the commented out `branches: "main"`). It would help to know pre-merge if the branch is not deployable.

internal_model_tool/common.py
```diff
@@ -0,0 +1,79 @@
...
+    hash = meta.get(f"model.versions.{version}")
+    return f"{version}.{hash}"
+
+
+def get_remote_meta(session, date):
+    sm = session.client("secretsmanager")
+    host = sm.get_secret_value(SecretId="elasticsearch/catalogue_api/public_host")[
+        "SecretString"
+    ]
+    username = sm.get_secret_value(
```
kenoir, line 47: You may want to create a user specifically for this tool (as with the diff_tool).

.buildkite/pipeline.yml
jamesgorrie: You're right about the assumption, but I wasn't sure, I'll leave it in and if it is hassle-some we can work it out then.

kenoir (approved):
Nice. 

Also potentially adds a  place to run the tool that would come from: https://github.com/wellcomecollection/platform/issues/5239

I can imagine if the compatibility test suggested in that issue passes, automatically adding the referenced internal model to the index meta on a merge to main.

PR #206, opened 2021-07-27
Standardise Catalogue API apps

kenoir:
In order to be less surprising for folk working on the Catalogue API make constructing services the same for items, requests and search.

This change adds model version checking to items & requests APIs and uses `WellcomeHttpApp` in the search app.


jamesgorrie (approved):
Real nice

PR #207, opened 2021-07-27
put user safely, without overriding password

jamesgorrie:
Ensures that if a user exists, we leave the password as is. 

What I might prefer over this is to [store the users as secrets in terraform](https://github.com/wellcomecollection/catalogue-api/pull/208) but I just want to have something blocking us from destroying things for now.

### DIFF
--- a/terraform/shared/scripts/users.py
+++ b/terraform/shared/scripts/users.py
@@ -1,0 +1,38 @@
+import elasticsearch
+import secrets
+
+"""
+This method checks if a user exists.
+Exists: it updates the roles if needed with no password update, returns None
+Missing: creates the user with a new password , returns username, password
+"""
+def put_user_safely(es, username, roles):
+    try:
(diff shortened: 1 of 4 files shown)


terraform/shared/scripts/users.py
```diff
@@ -0,0 +1,40 @@
...
+                    "roles": roles
+                }
+            )
+            print(f"user: {username} roles updated")
+        else:
+            print(f"user: {username} has no role updates")
+
+        return None
+
+    except elasticsearch.exceptions.NotFoundError:
```
alexwlchan, line 29: General good Python practice is to wrap `try … except` blocks as tight as possible – in this case, around the `data = es.security.get_user(…)` line, so if this is thrown unexpectedly elsewhere it still raises.

jamesgorrie:
Closing in favour of #208

PR #208, opened 2021-07-27
add test for provisioning users for elastic in terraform

jamesgorrie:
Allows us to provision users for elastic from terraform.

* Creates secret username & pass
* Creates secret value for username & pass
* Adds username & pass to elastic via `local-exec`

The `local-exec` script will never blat users, specifically their passwords.

Keeps it all nice and terraform-y.

Things to think about:

**updating a user**
This might be nice to keep as "needs more explicit input from human" as for many services is a breaking change. You could just have a script that `terraform apply -replace module.elastic-username`.


terraform/modules/elastic_user/create_user.py
```diff
@@ -0,0 +1,53 @@
...
+        return username, password
+
+
+if __name__ == '__main__':
+    try:
+        username = sys.argv[1]
+        password = sys.argv[2]
+        roles = sys.argv[3].split(",")
+    except IndexError:
+        sys.exit(f"Usage: {__file__} <USERNAME> <PASSWORD>")
```
jamesgorrie, line 51:
```suggestion
        sys.exit(f"Usage: {__file__} <USERNAME> <PASSWORD> <ROLES>")
```

kenoir:
Nice idea.

terraform/modules/elastic_user/main.tf
```diff
@@ -0,0 +1,37 @@
...
+  secret_id     = aws_secretsmanager_secret.username.id
+  secret_string = var.service
+}
+
+resource "aws_secretsmanager_secret_version" "password" {
+  secret_id     = aws_secretsmanager_secret.username.id
+  secret_string = random_password.password.result
+
+  provisioner "local-exec" {
+    command = "python3 ../modules/elastic_user/create_user.py ${var.service} ${random_password.password.result} ${join(",", var.roles)}"
```
kenoir, line 29: Will the roles get updated when the var changes? I'm not sure that a change in `var.roles` would retrigger this provisioner.

jamesgorrie: Nope, that's a piece I am trying to work out. In the description I am thinking we could have an `update_user` script, but would prefer to have it happen automatically.

Maybe a `null_resource` or something with that as a dep that runs an update_user_roles script... I'll have a play.

kenoir (approved):

kenoir:
Are you going to import existing users described by the script in a future PR?

jamesgorrie:
> Are you going to import existing users described by the script in a future PR?

Definitely - it'll be a bit of a faff, but we shouldn't have two ways of doing a sensitive thing laying about.

PR #212, opened 2021-07-29
move more from scripts into terraform for elastic

jamesgorrie:
* Adds elastic users to readme
* adds elastic settings to terraform config rather than in a script (we can deprecate this soon)
* only use script to create roles


terraform/shared/elastic.tf
```diff
@@ -17,6 +17,18 @@ resource "ec_deployment" "catalogue_api" {
...
     }
+
+    config {
+      user_settings_yaml = yamlencode(
+        {
+          # This setting is a fix for performance problems with an optimisation added in 7.13.1
+          # See: https://www.elastic.co/guide/en/elasticsearch/reference/7.13/release-notes-7.13.2.html
+          # "Add setting to disable aggs optimization #73620 (issue: #73426)"
+          # TODO: A fix is likely in ES 7.14, after which we can remove this setting.
+          "search.aggs.rewrite_to_filter_by_filter" : true
```
jamesgorrie, line 28: This won't make a difference really as:

> The order of precedence for cluster settings is:
>    * transient cluster settings
>    * persistent cluster settings
>    * settings in the elasticsearch.yml configuration file.

-- https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-update-settings.html#cluster-update-settings-api-desc

kenoir (approved):

PR #213, opened 2021-07-30
Assorted Terraform house-cleaning and tidies

alexwlchan:
* Use a similar set of default tags to what we use everywhere else
* Remove some unused ECR repositories
* Use the shared secrets module (https://github.com/wellcomecollection/platform/issues/5249)
* Fix a warning from Terraform
* Remove some VPC endpoints for CloudWatch services that are unnecessary now we send stuff to Elasticsearch; cheaper to send it over the NAT Gateway than run a persistent VPC endpoint (which is costing us ~$5.67/day)


terraform/modules/stack/network.tf
```diff
@@ -1,7 +1,7 @@
-module "interface_endpoints" {
+/*module "interface_endpoints" {
   source = "../interface_endpoints"
 
   vpc_id            = var.vpc_id
   security_group_id = aws_security_group.egress.id
   subnet_ids        = var.private_subnets
-}
+}*/
```
jamesgorrie, line 7: Should we just remove this? If we're keeping it, what for?

jamesgorrie (approved):

terraform/modules/stack/network.tf
alexwlchan: Yeah, it should be removed. Done.

PR #216, opened 2021-08-09
Tidy up the construction of SierraService instances

alexwlchan:
This moves the SierraService fixture into the common library and shares it among the apps. Partly because it simplifies the tests, and partly because it reduces the number of places where we create new instances of SierraService – which is useful for https://github.com/wellcomecollection/platform/issues/5255, which I’m looking at next.


alexwlchan:
This has got a bit messy and out of hand. Let's wait for other PRs to merge and then I'll redo this.

kenoir (approved):

Issue #223, opened 2021-08-12
Return Item titles from Requests API


kenoir:
The requests API should return the item title when asked to list a users holds.

This is to enable the front-end to display these titles alongside a hold without having to make a 2nd request to discover this information, reducing time to render to render the page and simplifying code in the front-end.

PR #224, opened 2021-08-12
Provide filled out item data when returning hold list

kenoir:
Put the full item into a holds list for the front-end.

Closes https://github.com/wellcomecollection/catalogue-api/issues/223

### DIFF
--- a/common/stacks/src/main/scala/weco/api/stacks/models/display/DisplayResultsList.scala
+++ b/common/stacks/src/main/scala/weco/api/stacks/models/display/DisplayResultsList.scala
@@ -28,7 +28,8 @@ case class DisplayResultsList(
 object DisplayRequest {
   def apply(hold: StacksHold): DisplayRequest = {
     DisplayRequest(
-      // TODO: Remove this .get!
+      // TODO: This .get should always be Some here
+      // TODO: but we should refactor to remove it!
       item = DisplayItem(
         item = hold.item.get,
         includesIdentifiers = true
--- a/common/stacks/src/main/scala/weco/api/stacks/models/display/DisplayResultsList.scala
+++ b/common/stacks/src/main/scala/weco/api/stacks/models/display/DisplayResultsList.scala
@@ -28,9 +28,10 @@ case class DisplayResultsList(
 object DisplayRequest {
   def apply(hold: StacksHold): DisplayRequest = {
     DisplayRequest(
-      item = new DisplayItem(
-        id = hold.canonicalId.map { _.underlying },
-        identifiers = Some(List(DisplayIdentifier(hold.sourceIdentifier)))
+      // TODO: Remove this .get!
+      item = DisplayItem(
+        item = hold.item.get,
+        includesIdentifiers = true
       ),
       pickupDate = hold.pickup.pickUpBy,
       pickupLocation = DisplayLocationDescription(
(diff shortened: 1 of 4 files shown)


requests/src/main/scala/weco/api/requests/responses/LookupPendingRequests.scala
```diff
@@ -21,16 +25,16 @@ trait LookupPendingRequests extends CustomDirectives {
...
 
         holdsWithCatalogueIds <- Future.sequence(
-          userHolds.right.get.holds.map { hold =>
+          userHolds.right.get.holds.map { hold: StacksHold =>
             itemLookup
               .bySourceIdentifier(hold.sourceIdentifier)
               .map {
-                case Left(elasticError) =>
-                  warn(s"Unable to look up $hold in Elasticsearch")
+                case Left(elasticError: ElasticsearchError) =>
```
kenoir, line 32: Previously we just chucked this error, not sure why - i've added it to the warning here.

common/stacks/src/main/scala/weco/api/stacks/models/display/DisplayResultsList.scala
```diff
@@ -28,9 +28,11 @@ case class DisplayResultsList(
 object DisplayRequest {
   def apply(hold: StacksHold): DisplayRequest = {
     DisplayRequest(
-      item = new DisplayItem(
-        id = hold.canonicalId.map { _.underlying },
-        identifiers = Some(List(DisplayIdentifier(hold.sourceIdentifier)))
+      // TODO: This .get should always be Some here
```
kenoir, line 31: I'll refactor this next, but to speed up getting this data into the front-end we could merge this PR, deploy and put the refactor into a new PR.

requests/src/main/scala/weco/api/requests/responses/LookupPendingRequests.scala
```diff
@@ -21,16 +25,18 @@ trait LookupPendingRequests extends CustomDirectives {
...
+                  warn(
+                    s"Unable to look up $hold in Elasticsearch: ${elasticError}"
```
alexwlchan, line 34: Let's not address it here for the sake of getting this PR out quickly, but having a sourceIdentifier on the hold feels a bit odd to me – a hold is a thing in Sierra and could have a source identifier, it's just not the identifier we're using here.

alexwlchan: What will happen in this branch if we can't find the item?

requests/src/test/scala/weco/api/requests/RequestsApiFeatureTest.scala
```diff
@@ -98,6 +102,7 @@ class RequestsApiFeatureTest
              |            "type" : "Identifier"
              |          }
              |        ],
+             |        "title" : "${titleString}",
```
alexwlchan, line 105: nit:

```suggestion
             |        "title" : "$titleString",
```

alexwlchan (approved):

requests/src/main/scala/weco/api/requests/responses/LookupPendingRequests.scala
kenoir: > but having a sourceIdentifier on the hold feels a bit odd to me

I agree, we use it to look up the item later and fill the Item option on `StacksHold` which itself feels wrong. I expect that it'll get removed refactoring out the `.get`.

> What will happen in this branch if we can't find the item?

Good point. I think this is actually a case that will blow up the `.get` on the option added in this PR (the`StacksHold` is returned with a `None` for an `Item`. It's still quite unlikely but we should handle this by doing a multi-item get and returning a list of the items we can get while logging that things are missing with a `warn`.

PR #228, opened 2021-08-13
Simplify the requests API

kenoir:
Folllowing https://github.com/wellcomecollection/catalogue-api/pull/224

Refactors the requests API to remove some potential bugs.


common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
```diff
@@ -104,16 +104,11 @@ class SierraService(
...
-
-          case Right(holds) if holds.holds.size >= holdLimit =>
+        getHolds(patron).flatMap {
+          case holds if holds.size >= holdLimit =>
             Future.successful(Left(HoldRejected.UserIsAtHoldLimit))
-
+          case holds if holds.keys.toList.contains(sourceIdentifier) =>
+            Future.successful(Right(HoldAccepted.HoldAlreadyExists))
           case _ => checkIfItemCanBeRequested(item)
         }
```
alexwlchan, line 113: I think you've introduced a bug here.

Consider: what if the user has a hold on this item *and* the user is at their hold limit?

common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
```diff
@@ -235,44 +230,22 @@ class SierraService(
...
-            )
-          )
+  def getHolds(patronNumber: SierraPatronNumber) =
+    for {
+      holds <- sierraSource
+        .listHolds(patronNumber)
+        .flatMap {
+          case Right(holds) => Future(holds)
+          // TODO: Proper error here please
+          case Left(err) => Future.failed(new Throwable(s"$err"))
```
alexwlchan: Are you going to do the proper error in this PR?

common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
```diff
@@ -235,44 +230,22 @@ class SierraService(
...
-              userId = patronNumber.withoutCheckDigit,
-              holds = holds.entries.map(buildStacksHold)
-            )
-          )
+  def getHolds(patronNumber: SierraPatronNumber) =
+    for {
+      holds <- sierraSource
+        .listHolds(patronNumber)
+        .flatMap {
+          case Right(holds) => Future(holds)
```
alexwlchan, line 238:
```suggestion
          case Right(holds) => Future.successful(holds)
```

Let's be explicit and match the left arm.

common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
```diff
@@ -235,44 +230,22 @@ class SierraService(
...
+          case Left(err) => Future.failed(new Throwable(s"$err"))
+        }
 
-        case Left(err) => Left(err)
+      sourceIdentifiers = holds.entries.map { hold =>
+        SierraItemIdentifier.toSourceIdentifier(
+          SierraItemIdentifier.fromUrl(hold.record)
+        )
       }
+    } yield sourceIdentifiers.zip(holds.entries).toMap
```
alexwlchan, line 252: Minor: do we need to iterate over `holds.entries` twice? I find something like this a bit clearer, possibly because I can never remember what order `zip` goes in:

```suggestion
      sourceIdentifiers = holds.entries.map { hold =>
        val identifier = SierraItemIdentifier.toSourceIdentifier(
          SierraItemIdentifier.fromUrl(hold.record)
        )
        
        identifier -> hold
      }
    } yield sourceIdentifiers.toMap
```

common/stacks/src/test/scala/weco/api/stacks/services/SierraServiceTest.scala
```diff
@@ -239,216 +239,174 @@ class SierraServiceTest
...
-                    id = "i",
-                    label = "item hold ready for pickup"
-                  )
-                )
-              )
-            )
-          }
+          //          val future = service.getStacksUserHolds(patron)
+          //          whenReady(future) {
+          true shouldBe false
```
alexwlchan: ?

common/stacks/src/test/scala/weco/api/stacks/services/SierraServiceTest.scala
```diff
@@ -239,216 +239,174 @@ class SierraServiceTest
...
+        val future = service.placeHold(
+          patron = patron,
+          sourceIdentifier = sourceIdentifier
+        )
 
-          whenReady(future) {
-            _.left.value shouldBe HoldRejected.ItemCannotBeRequested
-          }
+        whenReady(future) {
+          _.left.value shouldBe HoldRejected.ItemCannotBeRequested
```
alexwlchan, line 406: I'm sort of trusting you haven't fiddled with this file too much, because the indentation makes it harder to tell.

requests/src/main/scala/weco/api/requests/responses/LookupPendingRequests.scala
```diff
@@ -20,31 +16,38 @@ trait LookupPendingRequests extends CustomDirectives {
...
+      holdsMap <- sierraService.getHolds(patronNumber)
+
+      itemLookupResults <- Future
+        .sequence(
+          holdsMap.keys
+            .map(itemLookup.bySourceIdentifier)
+            .map(_.map {
+              case Right(item) => Some(item)
+              case Left(err) =>
+                warn(s"$err")
```
alexwlchan: We should put some context on these warnings, so we can more easily trace them to a line of code.

Do we get the source identifier we were trying to look up?

common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
kenoir: 👍  re-ordering the case statements will fix that I think. I don't think we have a test for that particular case.

common/stacks/src/test/scala/weco/api/stacks/services/SierraServiceTest.scala
kenoir: review draft PRs at your peril!

common/stacks/src/main/scala/weco/api/stacks/services/SierraService.scala
kenoir: That is nicer 👍

common/stacks/src/test/scala/weco/api/stacks/services/SierraServiceTest.scala
kenoir: I went through and de-jiggled it - I think a curly bracket had tried to make a run for it.

kenoir:
I think this could do with some more tests around the failure cases on `lookupRequests`, but might be for a future PR - depending on review.

requests/src/main/scala/weco/api/requests/responses/LookupPendingRequests.scala
```diff
@@ -20,31 +17,45 @@ trait LookupPendingRequests extends CustomDirectives {
...
-          }
+    val itemHolds = for {
+      holdsMap <- sierraService.getHolds(patronNumber)
+
+      itemLookupResults <- Future
+        .sequence(
+          holdsMap.keys
+            .map(srcId => (srcId, itemLookup.bySourceIdentifier(srcId)))
+            .map {
+              case (srcId, itemLookup) =>
```
alexwlchan, line 28: You're reusing the variable name `itemLookup`, which feels like a potential source of confusion.

requests/src/main/scala/weco/api/requests/responses/LookupPendingRequests.scala
```diff
@@ -20,31 +17,45 @@ trait LookupPendingRequests extends CustomDirectives {
...
-                  hold.copy(item = Some(item))
-              }
-          }
+    val itemHolds = for {
+      holdsMap <- sierraService.getHolds(patronNumber)
+
+      itemLookupResults <- Future
+        .sequence(
+          holdsMap.keys
+            .map(srcId => (srcId, itemLookup.bySourceIdentifier(srcId)))
```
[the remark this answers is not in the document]
kenoir: I think this doesn't change existing behaviour, so i'd be reluctant to refactor in this PR. I think you're right though, it needs addressing. I'd prefer to tackle that in a new PR.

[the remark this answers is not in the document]
alexwlchan: Agreed, let’s do that separately.

alexwlchan (approved):
🥡

PR #230, opened 2021-08-16
Adds executeMultiSearchRequest to ElasticsearchService

kenoir:
Follows https://github.com/wellcomecollection/catalogue-api/pull/228

Addresses performance implications of making many requests to Elasticsearch to retrieve item data raised here: https://github.com/wellcomecollection/catalogue-api/pull/228#discussion_r688886503

Also adds some tests for making single search requests that were missing for `ElasticsearchService`.

### DIFF
--- a/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
+++ b/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
@@ -3,7 +3,7 @@ package weco.api.search.elasticsearch
 import com.sksamuel.elastic4s.ElasticDsl._
 import com.sksamuel.elastic4s.circe._
 import com.sksamuel.elastic4s.requests.get.GetResponse
-import com.sksamuel.elastic4s.requests.searches.{SearchRequest, SearchResponse}
+import com.sksamuel.elastic4s.requests.searches.{MultiSearchRequest, MultiSearchResponse, SearchRequest, SearchResponse}
 import com.sksamuel.elastic4s.{ElasticClient, Hit, Index, Response}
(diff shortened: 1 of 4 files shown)


common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
```diff
@@ -72,6 +77,37 @@ class ElasticsearchService(elasticClient: ElasticClient)(
         }
     }
 
+  def executeMultiSearchRequest(
+    request: MultiSearchRequest
+  ): Future[Either[ElasticsearchError, MultiSearchResponse]] =
+    spanFuture(
+      name = "ElasticSearch#executeMultiSearchRequest",
```
kenoir, line 84: @jamieparkinson i'm wondering if this is the right way to do this? I'm guessing this will distinguish multiSearchRequests in APM.

The other thing is that i'm accumulating total time to return all the requests in a multi-request, when maybe we want both accumulated and individual time?

jamieparkinson: Think this is right - to be honest it might be worth revisiting the tracing stuff as I did it a long time ago and I think the [state](https://github.com/elastic/apm-agent-java/pull/1048) of scala support has improved a bit since then.

Anyway, re the timing, I would agree that we should capture the individual times (maybe only those, as the total is just derived from those, and the total is maybe a bit misleading as [I think?] the requests can run in parallel).

iirc you can set whatever you want as keys in `transaction.setLabel`; I don't think `elasticTook` isn't special or defined elsewhere.

jamieparkinson (approved):

Issue #231, opened 2021-08-16
Improve Requests API design


kenoir:
The requests API when serving a list of holds for a user must query the Sierra API then extract item data for each hold from the Catalogue API.

In order that the code to serve a list of users holds is robust and performant we should perform the following refactors:
- [x] Simplify code around retrieving holds and ensure it is robust (https://github.com/wellcomecollection/catalogue-api/pull/228)
- Address performance issues of multiple catalogue API requests:
  - [x] Add the ability to perform [multiSearchRequests](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-multi-search.html) (https://github.com/wellcomecollection/catalogue-api/pull/230)
  - [x] Integrate multi-search requests into requests API

PR #235, opened 2021-08-17
Adds findByMultiSearch to Elasticsearchservice

kenoir:
Part of https://github.com/wellcomecollection/catalogue-api/issues/231

In order to eventually update [`ItemLookup`](https://github.com/wellcomecollection/catalogue-api/blob/main/common/stacks/src/main/scala/weco/api/stacks/services/ItemLookup.scala#L13) to query on multiple `SourceIdentifier` in a performant way this change adds a `findByMultiSearch` function to `ElasticsearchService`.

This change also add more tests `ElasticsearchService`. It is used in some fundamental operations and is quite under-tested.

### DIFF
--- a/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
+++ b/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
@@ -3,12 +3,7 @@ package weco.api.search.elasticsearch
 import com.sksamuel.elastic4s.ElasticDsl._
 import com.sksamuel.elastic4s.circe._
 import com.sksamuel.elastic4s.requests.get.GetResponse
-import com.sksamuel.elastic4s.requests.searches.{
-  MultiSearchRequest,
-  MultiSearchResponse,
-  SearchRequest,
-  SearchResponse
-}
+import com.sksamuel.elastic4s.requests.searches.{MultiSearchRequest, SearchRequest, SearchResponse}
 import com.sksamuel.elastic4s.{ElasticClient, Hit, Index, Response}
 import grizzled.slf4j.Logging
 import io.circe.Decoder
@@ -57,23 +52,14 @@ class ElasticsearchService(elasticClient: ElasticClient)(
   def findByMultiSearch[T](
                        request: MultiSearchRequest
   )(implicit decoder: Decoder[T]): Future[(Seq[ElasticsearchError], Seq[T])] =
-    for {
-      response <- executeMultiSearchRequest(request)
+    executeMultiSearchRequest(request).map {
+      case (errors, responses) =>
+        val deserialisedResponses = responses
+          .flatMap(_.hits.hits)
+          .map(deserialize[T])
 
-      results = response match {
-        case Right(searchResponse: MultiSearchResponse) =>
-          val failures = searchResponse.failures
-            .map(ElasticsearchError(_))
-
-          val successes = searchResponse.successes
-            .flatMap(_.hits.hits)
-            .map(deserialize[T])
-
-          (failures, successes)
-
-        case Left(err) => (Seq(err), List.empty)
-      }
-    } yield results
+        (errors, deserialisedResponses)
+    }
 
   def executeSearchRequest(
     request: SearchRequest
@@ -100,7 +86,7 @@ class ElasticsearchService(elasticClient: ElasticClient)(
 
   def executeMultiSearchRequest(
     request: MultiSearchRequest
-  ): Future[Either[ElasticsearchError, MultiSearchResponse]] =
+  ): Future[(Seq[ElasticsearchError], Seq[SearchResponse])] =
     spanFuture(
       name = "ElasticSearch#executeMultiSearchRequest",
       spanType = "request",
@@ -113,27 +99,37 @@ class ElasticsearchService(elasticClient: ElasticClient)(
         .map(_.toEither)
         .map {
           case Right(multiResponse) =>
-            val accumulatedTime =
-              multiResponse.items.zipWithIndex.foldLeft(0L) {
-                (acc, itemWithIndex) =>
-                  val (item, index) = itemWithIndex
+            val foldInitial = (0L, Seq.empty[Long], Seq.empty[ElasticsearchError], Seq.empty[SearchResponse])
+
+            val (finalTotalTimeTaken, finalTimesTaken, finalErrors, finalSearchResponses) =
+              multiResponse.items.foldLeft(foldInitial) {
+                (acc, item) =>
+                  val (timeTakenTotal, timesTaken, errors, searchResponses) = acc
 
                   item.response match {
-                    case Right(itemResponse) =>
-                      transaction.setLabel(
-                        s"elasticTook-${index}",
-                        itemResponse.took
-                      )
-                      acc + itemResponse.took
-                    case Left(_) => acc
+                    case Right(itemResponse) => {
+                      val updatedTotalTimeTaken = timeTakenTotal + itemResponse.took
+                      val updatedTimesTaken = timesTaken :+ itemResponse.took
+                      val updatedSearchResponses = searchResponses :+ itemResponse
+
+                      (updatedTotalTimeTaken, updatedTimesTaken, errors, updatedSearchResponses)
+                    }
+
+                    case Left(error) => {
+                      val updatedErrors = errors :+ ElasticsearchError(error)
+                      (timeTakenTotal, timesTaken, updatedErrors, searchResponses)
+                    }
                   }
               }
 
-            transaction.setLabel("elasticTookTotal", accumulatedTime)
-            Right(multiResponse)
+            finalTimesTaken.zipWithIndex.map { case (timeTaken, index) =>
+              transaction.setLabel(s"elasticTook-$index", timeTaken)
+            }
+            transaction.setLabel("elasticTookTotal", finalTotalTimeTaken)
 
-          case Left(err) =>
-            Left(ElasticsearchError(err))
+            (finalErrors, finalSearchResponses)
+
+          case Left(err) => (Seq(ElasticsearchError(err)), Seq.empty)
         }
     }
 
--- a/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
+++ b/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
@@ -3,7 +3,11 @@ package weco.api.search.elasticsearch
 import com.sksamuel.elastic4s.ElasticDsl._
 import com.sksamuel.elastic4s.circe._
 import com.sksamuel.elastic4s.requests.get.GetResponse
-import com.sksamuel.elastic4s.requests.searches.{MultiSearchRequest, SearchRequest, SearchResponse}
+import com.sksamuel.elastic4s.requests.searches.{
+  MultiSearchRequest,
+  SearchRequest,
+  SearchResponse
+}
 import com.sksamuel.elastic4s.{ElasticClient, Hit, Index, Response}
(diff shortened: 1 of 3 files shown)


common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -36,14 +40,14 @@ class ElasticsearchServiceTest
     with RandomGenerators
     with ElasticsearchFixtures {
 
-  case class ExampleThing(id: String, name: String)
+  case class ExampleThing(id: CanonicalId, name: String)
 
-  def randomThing = ExampleThing(
-    id = randomAlphanumeric(10).toLowerCase,
+  def randomThing: ExampleThing = ExampleThing(
+    id = CanonicalId(randomAlphanumeric(8).toLowerCase),
```
alexwlchan, line 46: minor: we have IdentifierGenerators available for this sort of thing.

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -82,15 +86,185 @@ class ElasticsearchServiceTest
...
+          val foundById = findByIdEither.right.value
+          foundById shouldBe thingToQueryFor
+        }
+      }
+    }
+
+    it("should return an appropriate error if no ID matches") {
+      val thingsToIndex = 0.to(randomInt(5, 10)).map(_ => randomThing).toList
+
+      withExampleIndex(thingsToIndex) { index =>
```
alexwlchan, line 110: minor: I would skip indexing anything here, and go straight to querying.

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -82,15 +86,185 @@ class ElasticsearchServiceTest
...
+
+          failedFindByID shouldBe a[DocumentNotFoundError[_]]
+        }
+      }
+    }
+
+    it("should return an appropriate error if the index does not exist") {
+      val thingsToIndex = 0.to(randomInt(5, 10)).map(_ => randomThing).toList
+
+      withExampleIndex(thingsToIndex) { _ =>
```
alexwlchan, line 128: Similar, skip creating an index at all if we're not going to use it.

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -82,15 +86,185 @@ class ElasticsearchServiceTest
...
+        val elasticsearchService = new ElasticsearchService(elasticClient)
+        val thingToQueryFor = thingsToIndex.head
+        val badIndex = Index(randomAlphanumeric(10))
```
alexwlchan, line 131: we have helpers for this:

```suggestion
        val badIndex = createIndex
```

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -101,13 +275,46 @@ class ElasticsearchServiceTest
...
           elasticsearchService.executeMultiSearchRequest(multiSearchRequest)
 
-        whenReady(multiSearchResponseFuture) { multiSearchResponseEither =>
-          val multiSearchResponse = multiSearchResponseEither.right.value
-          val queryResults =
-            multiSearchResponse.items.map(_.response.right.value)
-          val returnedThings = queryResults.flatMap(_.to[ExampleThing])
+        whenReady(multiSearchResponseFuture) {
+          case (errors, searchResponses) =>
+            errors should have size (0)
```
alexwlchan, line 280:
```suggestion
            errors should be empty
```

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -101,13 +275,46 @@ class ElasticsearchServiceTest
...
+
+            val returnedThings = searchResponses.flatMap(_.to[ExampleThing])
+            returnedThings.toSet shouldBe thingsToQueryFor.toSet
+        }
+      }
+    }
+
+    it("returns an appropriate error when the specified index does not exist") {
```
alexwlchan, line 288: this test description feels a bit off to me – you're testing that you get an error *and* that you get the results that work

alexwlchan (commented):
One more thought: is it worth implementing executeSearchRequest in terms of executeMultiSearch? Wrap the single request in a MultiSearchRequest, unpack the response?

kenoir:
> One more thought: is it worth implementing executeSearchRequest in terms of executeMultiSearch? Wrap the single request in a MultiSearchRequest, unpack the response?

 I think when we call executeSearchRequest we don't want to invoke multiSearch with whatever differences that entails in the underlying implementation.

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
kenoir: `shouldBe empty` upset intellij syntax highlighting ... but it compiles fine.

common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
```diff
@@ -92,27 +103,52 @@ class ElasticsearchService(elasticClient: ElasticClient)(
...
+            val foldInitial = (
+              0L,
+              Seq.empty[Long],
+              Seq.empty[ElasticsearchError],
+              Seq.empty[SearchResponse]
+            )
+
+            val (
+              finalTotalTimeTaken,
+              finalTimesTaken,
```
alexwlchan, line 115: I'm not sure I love having the total and the list of times tracked separately – feels like an opportunity for them to get out of sync. Could we just track the list and do a `.sum` on the list at the end?

I also wonder if there's a more time-y type we can use rather than `Long`, but I don't think we get that from elastic4s?

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -34,16 +39,17 @@ class ElasticsearchServiceTest
...
+    with IdentifiersGenerators
     with ElasticsearchFixtures {
 
-  case class ExampleThing(id: String, name: String)
+  case class ExampleThing(id: CanonicalId, name: String)
 
-  def randomThing = ExampleThing(
-    id = randomAlphanumeric(10).toLowerCase,
+  def randomThing: ExampleThing = ExampleThing(
+    id = createCanonicalId,
```
alexwlchan, line 48: 👍

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -82,15 +88,176 @@ class ElasticsearchServiceTest
...
+
+        val findByIdFuture =
+          elasticsearchService.findById[ExampleThing](badId)(index)
+
+        whenReady(findByIdFuture) { findByIdEither =>
+          val failedFindByID = findByIdEither.left.value
+
+          failedFindByID shouldBe a[DocumentNotFoundError[_]]
+        }
+      }
```
alexwlchan, line 124: Personally I find it nicer when I can bin the intermediate variables, but it's not super important:

```suggestion
        whenReady(findByIdFuture) {
          _.left.value shouldBe a[DocumentNotFoundError[_]]
        }
      }
```

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -82,15 +88,176 @@ class ElasticsearchServiceTest
...
+          name = thingToQueryFor.name
+        )
+
+        val findBySearchFuture =
+          elasticsearchService.findBySearch[ExampleThing](searchRequest)
+
+        whenReady(findBySearchFuture) { findBySearchEither =>
+          val foundBySearch = findBySearchEither.right.value
+          foundBySearch should have size (1)
+          foundBySearch.head shouldBe thingToQueryFor
```
alexwlchan, line 162: nit:

```suggestion
          foundBySearch shouldBe Seq(thingToQueryFor)
```

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -82,15 +88,176 @@ class ElasticsearchServiceTest
...
+        whenReady(findByMultiSearchFuture) {
+          case (errors, foundBySearch) =>
+            errors should have length (0)
+            foundBySearch.toSet shouldBe thingsToQueryFor.toSet
+        }
+      }
+    }
+
+    it(
+      "returns an appropriate error for only those queries where the specified index does not exist"
```
alexwlchan, line 215:
```suggestion
      "returns errors for queries that fail"
```

?

bit less of a mouthful?

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -82,15 +88,176 @@ class ElasticsearchServiceTest
...
+
+        whenReady(findByMultiSearchFuture) {
+          case (errors, foundBySearch) =>
+            foundBySearch.toSet shouldBe thingsToQueryFor.toSet
+
+            errors should have size (1)
+            errors.head shouldBe a[IndexNotFoundError]
+        }
+      }
+    }
```
alexwlchan, line 248: Do we need a test for the case where you call findMultiSearch and the data in the index doesn't match the model?

alexwlchan (approved):
A few small moments but nothing to block merging.

common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
kenoir: > I'm not sure I love having the total and the list of times tracked separately

The individual times and the total times are tracked within the same transaction so will be tied together by that.

Screenshot 2021-08-18 at 09 29 47

^ This is what a transaction looks like in the APM view of Kibana.

> Could we just track the list and do a .sum on the list at the end?

The summed time is interesting because it gives a view on how resource intensive the `multiSearchRequest` was, however it is not representative of the actual time taken as I believe Elasticsearch can execute the constituent queries in parallel. 

An example of where we may want to view the times for individual queries is to distinguish within a multiple request whether a single query within that group is taking a long time to complete.

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
kenoir: I think that's useful for all the test cases here. I'll add that in a future PR.

common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
alexwlchan: I should have been clearer – I’m fine with them being two individual metrics in Elasticsearch, it just seems odd that we’re managing them as two separate variables in this file.

kenoir: Ah sorry. The variable setup is to avoid side-effecting in a fold.

PR #238, opened 2021-08-18
Use findByMultiSearch in ItemLookup & Requests API

kenoir:
Continuing to work on making sensible catalogue API queries to fill out items on holds.

This change connects the requests API hold listing to the `findByMultiSearch` function through `ItemLookup`. Updates to the interface of `findByMultiSearch` were required to effectively identify missing items.

I've refrained from adding many more tests to requests or starting on the logic to add the correct work title to holds (see https://github.com/wellcomecollection/platform/issues/5267), that work will be required in future PRs.

Follows: https://github.com/wellcomecollection/catalogue-api/pull/235
Part of: https://github.com/wellcomecollection/catalogue-api/issues/231

### DIFF
--- a/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
+++ b/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
@@ -105,6 +105,7 @@ class ElasticsearchService(elasticClient: ElasticClient)(
       withActiveTrace(elasticClient.execute(request))
         .map(_.toEither)
         .map {
+          case Left(err) => throw err.asException
           case Right(multiResponse) =>
             val foldInitial = (
               0L,
@@ -146,8 +147,6 @@ class ElasticsearchService(elasticClient: ElasticClient)(
             transaction.setLabel("elasticTookTotal", finalTotalTimeTaken)
 
             finalResults
-
-          case Left(err) => throw err.asException
         }
     }
 
--- a/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
+++ b/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
@@ -55,15 +55,16 @@ class ElasticsearchService(elasticClient: ElasticClient)(
 
   def findByMultiSearch[T](
     request: MultiSearchRequest
-  )(implicit decoder: Decoder[T]): Future[(Seq[ElasticsearchError], Seq[T])] =
-    executeMultiSearchRequest(request).map {
-      case (errors, responses) =>
-        val deserialisedResponses = responses
-          .flatMap(_.hits.hits)
-          .map(deserialize[T])
-
-        (errors, deserialisedResponses)
-    }
+  )(implicit decoder: Decoder[T]): Future[Seq[Either[ElasticsearchError, Seq[T]]]] =
+    for {
+      multiSearchResults <- executeMultiSearchRequest(request)
+      deserialisedResults = multiSearchResults.map {
+        case Right(searchResponse) => {
+          Right(searchResponse.hits.hits.map(deserialize[T]).toSeq)
+        }
+        case Left(err) => Left(err)
+      }
+    } yield deserialisedResults
 
   def executeSearchRequest(
     request: SearchRequest
@@ -90,7 +91,7 @@ class ElasticsearchService(elasticClient: ElasticClient)(
 
   def executeMultiSearchRequest(
     request: MultiSearchRequest
-  ): Future[(Seq[ElasticsearchError], Seq[SearchResponse])] =
+  ): Future[Seq[Either[ElasticsearchError, SearchResponse]]] =
     spanFuture(
       name = "ElasticSearch#executeMultiSearchRequest",
       spanType = "request",
@@ -106,36 +107,32 @@ class ElasticsearchService(elasticClient: ElasticClient)(
             val foldInitial = (
               0L,
               Seq.empty[Long],
-              Seq.empty[ElasticsearchError],
-              Seq.empty[SearchResponse]
+              Seq.empty[Either[ElasticsearchError, SearchResponse]],
             )
 
             val (
               finalTotalTimeTaken,
               finalTimesTaken,
-              finalErrors,
-              finalSearchResponses
+              finalResults
             ) =
               multiResponse.items.foldLeft(foldInitial) { (acc, item) =>
-                val (timeTakenTotal, timesTaken, errors, searchResponses) = acc
+                val (timeTakenTotal, timesTaken, results) = acc
 
                 item.response match {
                   case Right(itemResponse) => {
-                    val updatedTotalTimeTaken = timeTakenTotal + itemResponse.took
-                    val updatedTimesTaken = timesTaken :+ itemResponse.took
-                    val updatedSearchResponses = searchResponses :+ itemResponse
-
                     (
-                      updatedTotalTimeTaken,
-                      updatedTimesTaken,
-                      errors,
-                      updatedSearchResponses
+                      timeTakenTotal + itemResponse.took,
+                      timesTaken :+ itemResponse.took,
+                      results :+ Right(itemResponse)
                     )
                   }
 
                   case Left(error) => {
-                    val updatedErrors = errors :+ ElasticsearchError(error)
-                    (timeTakenTotal, timesTaken, updatedErrors, searchResponses)
+                    (
+                      timeTakenTotal,
+                      timesTaken,
+                      results :+ Left(ElasticsearchError(error))
+                    )
                   }
                 }
               }
@@ -146,9 +143,9 @@ class ElasticsearchService(elasticClient: ElasticClient)(
             }
             transaction.setLabel("elasticTookTotal", finalTotalTimeTaken)
 
-            (finalErrors, finalSearchResponses)
+            finalResults
 
-          case Left(err) => (Seq(ElasticsearchError(err)), Seq.empty)
+          case Left(err) => throw err.asException
         }
     }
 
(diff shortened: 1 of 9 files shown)


common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
```diff
@@ -146,9 +145,9 @@ class ElasticsearchService(elasticClient: ElasticClient)(
             }
             transaction.setLabel("elasticTookTotal", finalTotalTimeTaken)
 
-            (finalErrors, finalSearchResponses)
+            finalResults
 
-          case Left(err) => (Seq(ElasticsearchError(err)), Seq.empty)
+          case Left(err) => throw err.asException
```
alexwlchan, line 150: Is an unhandled throw safe here? We'll catch it and wrap it in a Future?

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -105,6 +115,24 @@ class ElasticsearchServiceTest
       }
     }
 
+    it("fails if deserialising to a different type than stored") {
+      val thingsToIndex = 0.to(randomInt(5, 10)).map(_ => randomThing).toList
```
alexwlchan, line 119: I'm not actually sure if it's clearer, but we have a helper for this sort of construction in scala-libs:

```suggestion
      val thingsToIndex = collectionOf(min = 5, max = 10) { randomThing }.toList
```

(Possibly if that returned a List[T]?)

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
```diff
@@ -195,52 +275,100 @@ class ElasticsearchServiceTest
...
           .findByMultiSearch[ExampleThing](multiSearchRequest)
 
-        whenReady(findByMultiSearchFuture) {
-          case (errors, foundBySearch) =>
-            errors should have length (0)
-            foundBySearch.toSet shouldBe thingsToQueryFor.toSet
+        whenReady(findByMultiSearchFuture) { results =>
+          val (errorEither, foundBySearchEither) = results.partition(_.isLeft)
+          val errors = errorEither.map(_.left.value)
+          val foundBySearch = foundBySearchEither.map(_.right.value)
```
alexwlchan, line 281: Could we use a collect here?

```suggestion
          val errors = results.collect { case Left(err) => err }
          val foundBySearch = result.collect { case Right(found) => found }
```

common/stacks/src/main/scala/weco/api/stacks/services/elastic/ElasticItemLookup.scala
```diff
@@ -64,44 +65,77 @@ class ElasticItemLookup(
...
+  ) =
+    works
+      .flatMap { _.data.items }
+      .collect {
+        case item @ Item(id @ IdState.Identified(_, _, _), _, _, _)
+            if id.sourceIdentifier == sourceIdentifier =>
+          // This .asInstanceOf[] is a no-op to help the compiler see what
+          // we can see by reading the code.
+          item.asInstanceOf[Item[IdState.Identified]]
+      }
```
alexwlchan, line 99: I'm struggling to wrap my head around these new methods – prefixing the `sourceIdentifier` variables with an `item` or `work` would help me, I think. My brain sees `work` and `sourceIdentifier` next to each other, and instinctively wants to associate the two, even though I think the source identifier is actually for the item?

common/stacks/src/main/scala/weco/api/stacks/services/elastic/ElasticItemLookup.scala
```diff
@@ -64,44 +65,77 @@ class ElasticItemLookup(
...
+            termQuery(
+              "data.items.id.sourceIdentifier.value",
+              sourceIdentifier.value
+            )
+          )
+      )
+      // TODO: How many things do we need to retrieve here?
+      .size(10)
+
+  val searchRequestOnIndex = buildSourceIdentifierSearchRequest(index)(_)
```
alexwlchan, line 85:
```suggestion
  private val searchRequestOnIndex = buildSourceIdentifierSearchRequest(index)(_)
```

common/stacks/src/main/scala/weco/api/stacks/services/elastic/ElasticItemLookup.scala
```diff
@@ -64,44 +65,77 @@ class ElasticItemLookup(
...
-                  if id.sourceIdentifier == sourceIdentifier =>
-                // This .asInstanceOf[] is a no-op to help the compiler see what
-                // we can see by reading the code.
-                item.asInstanceOf[Item[IdState.Identified]]
+        matchItemsOnWorksBySourceIdentifier(works, sourceIdentifier).toList match {
+          case item :: _ => Right(item)
+          case List()    => Left(DocumentNotFoundError(sourceIdentifier))
+        }
+    }
+  }
```
alexwlchan, line 114: \*deletes all my comments on this function*

If I checkout your branch locally and do "Find Usages", the only place the `bySourceIdentifier` method is used is in the tests. Let's bin it.

common/stacks/src/main/scala/weco/api/stacks/services/elastic/ElasticItemLookup.scala
```diff
@@ -64,44 +65,77 @@ class ElasticItemLookup(
...
+    )
+
+    elasticsearchService
+      .findByMultiSearch[Work[Indexed]](multiSearchRequest)
+      .map {
+        _.zip(sourceIdentifiers).map {
+          case (Right(works), srcId) =>
+            matchItemsOnWorksBySourceIdentifier(works, srcId).toList match {
+              // TODO: We can return multiple items from multiple works
+              // TODO: Apply better logic when picking an item!
```
alexwlchan, line 130: Useful comment! I was going to suggest you added one to the same line in `bySourceIdentifier`, before realising we weren't using that function anywhere and it was easier to bin. :D

common/stacks/src/test/scala/weco/api/stacks/services/ItemLookupTest.scala
```diff
@@ -109,6 +109,79 @@ class ItemLookupTest
     }
   }
 
+  describe("bySourceIdentifiers") {
```
alexwlchan, line 112: Not for this patch, but it occurs to me that if we're going to have a MemoryItemLookup, we should really have generic tests to ensure it behaves the same way as ElasticItemLookup.

common/stacks/src/test/scala/weco/api/stacks/services/ItemLookupTest.scala
```diff
@@ -109,6 +109,79 @@ class ItemLookupTest
...
+      val lookup = createLookup(index = createIndex)
+      val future =
+        lookup.bySourceIdentifiers(Seq(createSourceIdentifier))
+
+      whenReady(future) { results =>
+        results should have length (1)
+        results.head.left.value shouldBe a[IndexNotFoundError]
+      }
+    }
+  }
```
alexwlchan, line 183: Good test cases. 👍

requests/src/main/scala/weco/api/requests/responses/LookupPendingRequests.scala
```diff
@@ -20,40 +20,24 @@ trait LookupPendingRequests extends CustomDirectives {
...
-        )
-
-      itemHoldTuples = itemLookupResults.flatMap { item =>
+      itemLookupResults <- itemLookup.bySourceIdentifiers(holdsMap.keys.toSeq)
+
+      itemsFound = itemLookupResults.zip(holdsMap.keys).flatMap {
+        case (Right(item), _) => Some(item)
+        case (Left(elasticError: ElasticsearchError), srcId) =>
+          warn(
+            s"Error looking up item ${srcId}. Elasticsearch error: ${elasticError}"
```
alexwlchan, line 29:
```suggestion
            s"Error looking up item $srcId. Elasticsearch error: $elasticError"
```

alexwlchan (approved):
A few small suggestions and a suggestion for some code to delete, but nothing substantial. Looking good!

common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
kenoir: This is wrapped in the future started by `elasticClient.execute(request)` - it's difficult to see at the end of match statement. I'll move this to the top so it's easier to see this is safe.

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
kenoir: It doesn't return a `List` but I do think it's clearer.

common/search/src/test/scala/weco/api/search/elasticsearch/ElasticsearchServiceTest.scala
kenoir: Running over these with with a map extracts their value but also verifies their handedness (?) using `EitherValues`.

common/stacks/src/main/scala/weco/api/stacks/services/elastic/ElasticItemLookup.scala
kenoir: If we remove/replace `bySourceIdentifier` there is no need for this to be its own function, so that kind of solves the problem.

common/stacks/src/test/scala/weco/api/stacks/services/ItemLookupTest.scala
kenoir: Yes, I noticed that. Agree, will add in future PR.

PR #239, opened 2021-08-18
[DRAFT] Adds a title drawn from parent Work to an Item request

kenoir:
Follows: https://github.com/wellcomecollection/catalogue-api/pull/238

For: https://github.com/wellcomecollection/platform/issues/5267

**Draft PR Warning**

This change shows what the code to add a title from a work on an item request might look like.

**Note:** This will change the requests API and require front-end changes to pick up the added title value - it should be non-breaking with the existing front-end implementation.


kenoir:
Closing as dead-end.

PR #241, opened 2021-08-19
Requests service refactoring for "hold titles"

kenoir:
Some refactoring to make https://github.com/wellcomecollection/platform/issues/5267 easier.

This changes:
- Moves the `ItemLookup` logic into the requests service as it's the only place it is used (making accounting for special cases in that service less weird).
- Creates a `RequestsService` to better separate display/back-end service interactions.


common/stacks/src/main/scala/weco/api/stacks/models/HoldResponse.scala
```diff
@@ -17,6 +17,7 @@ object HoldRejected {
   case object ItemCannotBeRequested extends HoldRejected
   case object ItemIsOnHoldForAnotherUser extends HoldRejected
   case object ItemMissingFromSourceSystem extends HoldRejected
+  case object ItemUnavailableFromSourceSystem extends HoldRejected
```
alexwlchan, line 20: What's the difference between "missing" and "unavailable"?

requests/src/main/scala/weco/api/requests/services/RequestsService.scala
```diff
@@ -0,0 +1,80 @@
...
+          )
+          ) =>
+        sierraService.placeHold(
+          patron = patronNumber,
+          sourceIdentifier = srcId
+        )
+
+      case Right(sourceIdentifier) =>
+        warn(s"Cannot request from source: $itemId / $sourceIdentifier")
+        Future.successful(Left(ItemUnavailableFromSourceSystem))
```
alexwlchan, line 49: These right arms look like they have completely different types. Is this right? When would we get any sort of source identifier from itemLookup?

requests/src/main/scala/weco/api/requests/services/RequestsService.scala
```diff
@@ -0,0 +1,80 @@
...
+
+      itemLookupResults <- itemLookup.bySourceIdentifier(holdsMap.keys.toSeq)
+
+      itemsFound = itemLookupResults.zip(holdsMap.keys).flatMap {
+        case (Right(item), _) => Some(item)
+        case (Left(elasticError: ElasticsearchError), srcId) =>
+          warn(
+            s"Error looking up item $srcId. Elasticsearch error: $elasticError"
+          )
+          None
```
alexwlchan, line 69: Reading this for the third time or something, I suddenly wonder: could we get in a situation where we're omitting holds from a user's list? And so we show them a warning "you can't request this item, you're at hold limit N", but they can only see N–1 items.

Put another way: should this be a warning, or an error?

requests/src/main/scala/weco/api/requests/responses/LookupPendingRequests.scala
```diff
@@ -1,43 +1,23 @@
...
-
-      itemHoldTuples = itemsFound.flatMap { item =>
-        holdsMap.get(item.id.sourceIdentifier).map { hold =>
-          (hold, item)
-        }
-      }
-
-    } yield itemHoldTuples.toList
+      itemHoldTuples <- requestsService.getRequests(patronNumber)
+    } yield itemHoldTuples
```
alexwlchan, line 20: Do we still need a for comprehension, or could this become:

```scala
val itemHolds = requestsService.getRequests(patronNumber)
```

alexwlchan (approved):

common/stacks/src/main/scala/weco/api/stacks/models/HoldResponse.scala
kenoir: I've renamed this status to `SourceSystemNotSupported`, to denote a rejected attempt to place a hold on a source system we can't create holds for.

requests/src/main/scala/weco/api/requests/services/RequestsService.scala
kenoir: The autoformatting has unhelpfully rearranged this. It's pattern matching on the `IdentifierType` in `SourceIdentifier`, the same behaviour this logic had before - i'll put the `if` back as it is definitely clearer.

requests/src/main/scala/weco/api/requests/services/RequestsService.scala
kenoir: We could end up in that situation, I think you're right - this is an error.

PR #242, opened 2021-08-19
More requests/items service refactoring

kenoir:
We can further reduce what is available in the shared "stacks" library as most of the code now splits neatly into items or requests.

For https://github.com/wellcomecollection/catalogue-api/pull/241


requests/src/test/scala/weco/api/requests/services/SierraRequestsServiceTest.scala
```diff
@@ -30,150 +26,6 @@ class SierraServiceTest
     with SierraServiceFixture {
 
   describe("SierraService") {
-    describe("getAccessConditions") {
```
kenoir, line 33: These tests are replaced by the extra case here: https://github.com/wellcomecollection/catalogue-api/pull/242/files#diff-9ed61b263ed01f0a064581dfb2ce9ab7ecdaa54dc15a08b555c08fbf1b6e6cb7R239

alexwlchan (commented):
The SierraSource fixture has also been added in scala-libs in a separate PR – is it worth waiting until that's merged and used here before reviewing?

kenoir:
> The SierraSource fixture has also been added in scala-libs in a separate PR – is it worth waiting until that's merged and used here before reviewing?

I'll come back and do that in a different PR. I think this change is substantial enough.

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -62,6 +72,45 @@ class SierraItemUpdater(sierraService: SierraService)(
           .getOrElse(item)
     } toSeq
 
+  private def getAccessCondition(itemData: SierraItemData): AccessCondition = {
+    // The bib ID is used for debugging purposes; the bib status is only used
+    // for consistency checking. We can use placeholder data here.
```
alexwlchan, line 77: Probably not for fixing here, but it occurs to me that the Sierra API does give us a list of bib IDs in the item responses, and we could just pick one to use here.

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -62,6 +72,45 @@ class SierraItemUpdater(sierraService: SierraService)(
...
+    for {
+      itemEither <- sierraSource.lookupItemEntries(itemNumbers)
+
+      accessConditions = itemEither match {
+        case Right(SierraItemDataEntries(_, _, entries)) =>
+          entries.map(item => item.id -> getAccessCondition(item)).toMap
+        case Left(
+            SierraItemLookupError.MissingItems(missingItems, itemsReturned)
+            ) =>
+          warn(s"Item lookup missing items: ${missingItems}")
```
alexwlchan, line 105: nit

```suggestion
          warn(s"Item lookup missing items: $missingItems")
```

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -62,6 +72,45 @@ class SierraItemUpdater(sierraService: SierraService)(
...
+          itemsReturned.map(item => item.id -> getAccessCondition(item)).toMap
+        case Left(itemLookupError) =>
+          error(s"Item lookup failed: ${itemLookupError}")
```
alexwlchan, line 108: nit

```suggestion
          error(s"Item lookup failed: $itemLookupError")
```

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -70,10 +119,27 @@ class SierraItemUpdater(sierraService: SierraService)(
...
-      .getAccessConditions(itemMap.keys.toSeq)
-      .map(
-        updateAccessConditions(itemMap, _)
-      )
+    val accessConditions = for {
+      accessConditionsMap <- getAccessConditions(itemMap.keys.toSeq)
+
+      // It is possible for there to be a situation where Sierra does not know about
+      // an Item that is in the Catalogue API (e.g. it's been deleted but the change
+      // has not yet propagated. In that case it gets status "NotRequestable")
```
alexwlchan, line 127:
```suggestion
      // has not yet propagated. In that case it gets method "NotRequestable")
```

I would add here that this should be extremely rare in practice – items that are removed are usually suppressed in Sierra, then deleted after a long delay. (Do we handle suppressed items correctly?)

We should log a warning here so we can see if it's happening a lot.

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -70,10 +119,27 @@ class SierraItemUpdater(sierraService: SierraService)(
...
+          accessConditionsMap.keySet.contains(sierraItemNumber)
+      } keySet
+
+      missingItemsMap = missingItemsKeys
+        .map(
+          _ -> AccessCondition(
+            method = AccessMethod.NotRequestable,
+          status = None,
+          note = None
+        )
```
alexwlchan, line 139:
```suggestion
          _ -> AccessCondition(method = AccessMethod.NotRequestable)
```

alexwlchan (approved):

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
kenoir: I think it's time to add this config to `scalafmt` 🙂 !
```
rewrite.rules = [SortImports, RedundantBraces]
rewrite.redundantBraces.stringInterpolation = true
```

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
kenoir: Oh, that is useful.

PR #245, opened 2021-08-20
Use sort/filter query to get add Work title & id to requests listing

kenoir:
Currently the Requests API provides only Item titles as a means to identify a request/hold. Item titles by themselves are insufficient for users to identify requested things.

This change adds the work CanonicalId and the work title of a matching work to the API response (note: items can appear on multiple works, we arbitrarily pick the one with the lowest valued sierra identifier). 

**Note for anyone picking this up**: It may be necessary to update the [API gateway description of the expected requests output](https://github.com/wellcomecollection/identity/blob/main/models/item-request-list.json#L2).

For https://github.com/wellcomecollection/catalogue-api/pull/241


requests/src/main/scala/weco/api/requests/Main.scala
```diff
@@ -4,7 +4,7 @@ import akka.actor.ActorSystem
 import com.typesafe.config.Config
 import weco.Tracing
 import weco.api.requests.services.{
-  ElasticItemLookup,
```
kenoir, line 7: This change ditches the memory version of ItemLookup - implementing a version that worked with the `bySourceIdentifier` function when querying and sorting on works became unwieldy.

requests/src/main/scala/weco/api/requests/services/ItemLookup.scala
```diff
@@ -1,21 +1,146 @@
...
+                    workId = workId,
+                    workTitle = title,
+                    item = item.asInstanceOf[Item[IdState.Identified]]
+                  )
+              }
+              .toList match {
+              // We can return multiple items from multiple works
+              // The sortBy above returns us the lowest bibId work association
+              case List(item) => Right(item)
+              case item :: _  => Right(item)
```
jamieparkinson, line 128: Are these cases not the same? ie I think `List(item) == item :: Nil`

jamieparkinson (approved):
Main thing is that I think a general comment (probably on `ItemLookup`) putting this into context would be helpful - in particular, covering all the context of why we're picking the first results when we might have several

requests/src/main/scala/weco/api/requests/services/ItemLookup.scala
alexwlchan: I think this case is for the list that has more than one item.

alexwlchan:
I'll pick up adding the comment to this; Robert and I discussed it last week.

requests/src/main/scala/weco/api/requests/services/ItemLookup.scala
jamieparkinson: Right, so I think we just need the 2nd line (which will perform the work of the first too)? ie `case item :: _  => Right(item)`

alexwlchan: Yeah, I see what you mean. I'll change this.

jamieparkinson: "take the first item of a non-empty list"

alexwlchan:
@jamieparkinson I've added some more comments to this code – mind taking a look to see if the explanation makes sense?

jamieparkinson (approved):
Very clear, nice

PR #246, opened 2021-08-23
Retry a single timeout from Elasticsearch

alexwlchan:
The recent 5xx alarms from the Catalogue API are caused by single timeouts connecting to the Elasticsearch cluster. They don't seem to be problems with the individual request or downtime in the cluster, just a momentary flake.

This patch adds logic to retry a single timeout when we see it.

For more explanation, see:

* the code comment in ElasticsearchService
* the [RetryOps object in scala-libs](https://github.com/wellcomecollection/scala-libs/blob/966099678c0767f4277652d952780524623a9a8f/storage/src/main/scala/weco/storage/RetryOps.scala
) that the FutureRetryOps object and its tests are based on

### DIFF
--- a/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
+++ b/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchService.scala
@@ -178,7 +178,9 @@ class ElasticsearchService(elasticClient: ElasticClient)(
 
   private def isRetryable(t: Throwable): Boolean =
     t match {
-      case JavaClientExceptionWrapper(exc: java.net.ConnectException) if exc.getMessage.startsWith("Timeout connecting to") => true
+      case JavaClientExceptionWrapper(exc: java.net.ConnectException)
+          if exc.getMessage.startsWith("Timeout connecting to") =>
+        true
 
       case _ => false
     }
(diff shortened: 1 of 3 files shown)


jamesgorrie (commented):
Just wondering if this is the right approach? Might it be worth digging into what the issue connecting might be? This would firstly mean we are connecting twice to the server, and thus potentially increasing the latency for those requests?

Given the cluseter is managed, and we are definitely not squeezing it from a requsting POV, it feels like we might be able to address the systemic problem?

alexwlchan:
> Just wondering if this is the right approach? Might it be worth digging into what the issue connecting might be? This would firstly mean we are connecting twice to the server, and thus potentially increasing the latency for those requests?
>
> Given the cluseter is managed, and we are definitely not squeezing it from a requsting POV, it feels like we might be able to address the systemic problem?

Yeah, making the timeouts not happen would definitely be preferable, but I have no idea how to go about debugging that. There's nothing in the Elasticsearch metrics that looks untoward when these queries come in, I don't know where to find other logs, and the queries themselves don't seem unusual.

e.g. on Saturday we had a timeout for https://api.wellcomecollection.org/catalogue/v2/works?include=identifiers&query=1663783

on another we were looking up a single work https://api.wellcomecollection.org/catalogue/v2/works/p2ebzusp?include=identifiers,images,items,subjects,genres,contributors,production,notes,parts,partOf,precededBy,succeededBy,languages,holdings

Possibly this is something APM can shed light on?

jamesgorrie:
My concern is that we start calling twice, and for some reason that number increases and it becomes closer to the norm.

We don't seem to be sending logs to the logging cluster for the elasticsearch monitoring, we can investigate that. APM is monitoring the API service, which I don't think would capture why something timed out.

Looking there is no massive spikes here, but the search rate is calculated as 

> time it takes to execture searches divided by number of searches submitted

so single spikes will get lost in the averaging.

Screenshot 2021-08-23 at 11 43 59

@jamieparkinson do you think this might have something to do with the aggregations we've been talking about?

jamieparkinson:
Really no idea to be honest - it does seem like a good place to look, but then again these example timeout searches aren't using the aggregations...

jamesgorrie (commented):
Thinking about this, this is the type of thing that would be really good to be able to toggle. 

Do we think we could get the tidy changes in, then maybe have a config variable `retry_elastic_failures` so that we can flip it in TF and see if there's a difference?

alexwlchan:
> Thinking about this, this is the type of thing that would be really good to be able to toggle.
>
> Do we think we could get the tidy changes in, then maybe have a config variable retry_elastic_failures so that we can flip it in TF and see if there's a difference?

Sounds good! Although it won't be for a few days, this is behind requesting on my todo list.

alexwlchan:
Closing for now due to lack of interest.

PR #251, opened 2021-08-31
Add our standard API Gateway responses

alexwlchan:
e.g. https://api.wellcomecollection.org/catalogue/v2/works/a222zvge/items

This now returns an instance of our Error type rather than the default API Gateway response.


jamesgorrie (approved):

alexwlchan:
This uses a new module which is based on the responses in the storage service. At some point we should use it everywhere, but this'll do for now.

jamesgorrie:
[Confirmation here](https://api.wellcomecollection.org/catalogue/v2/works/a222zvge/items) which used to be 

```JSON
{ "message": "Forbidden" }
```

PR #252, opened 2021-09-01
Better search query types in rank

jamesgorrie:
TBD


jamesgorrie:
Work was being done better by @harrisonpim in another upcoming PR

PR #253, opened 2021-09-02
Remove requests deployment from catalogue-api build

kenoir:
In order to partition permissions in builds (catalogue-api builds **should not** reach into the identity account), this change will remove deployment into the identity AWS account from this build and move it elsewhere.

This repository will still build and push ECR images, though they will now go to the catalogue AWS account and be pulled into the identity account when deployment is triggered.

**Note:** This will pause requests-api deployment until it's configured in: https://github.com/wellcomecollection/identity. This change will be done next.

### DIFF
--- a/terraform/shared/ecr.tf
+++ b/terraform/shared/ecr.tf
@@ -7,6 +7,7 @@ resource "aws_ecr_repository" "items" {
 }
 
 # This is in the identity account as it is deployed there
+# TODO: Remove this when we are deploying it from identity
 resource "aws_ecr_repository" "requests" {
   provider = aws.identity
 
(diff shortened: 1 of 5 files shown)


terraform/shared/ecr.tf
```diff
@@ -6,10 +6,7 @@ resource "aws_ecr_repository" "items" {
   }
 }
 
-# This is in the identity account as it is deployed there
 resource "aws_ecr_repository" "requests" {
```
kenoir, line 9: I have manipulated the terraform state already to make this work. I've removed the identity ECR repo from TF state and will need to manually delete it later after we have demonstrated deployments work using the catalogue account ECR repo.

jamieparkinson (approved):
Maybe we should look into using ECR Public for our application images to avoid the cross-account dance

PR #267, opened 2021-09-09
Rank types

harrisonpim:
With apologies for another huge diff 😬 

This PR changes the pattern for fetching/building search templates. Rather than fetching/validating from a known, pre-generated set of search templates, they're built individually, as and when they're needed. 

Composing a search template no longer requires redundant information like the `namespace`, which can be inferred from elsewhere. Templates are now composed using an `env` (specifying the location from which to fetch the queries, `local` or `remote`), and an `index` (specifying the index in rank cluster to run that query against).

These changes also apply to interfaces like [/search](https://catalogue-api-rank-kczgq1czq-weco1.vercel.app/search), where the user need only specify an index and an env


rank/pages/index.tsx
```diff
@@ -1,30 +1,24 @@
...
-
-export const getServerSideProps: GetServerSideProps<Props> = async ({
-  query,
-}) => {
-  const tests = (
-    Array.isArray(query.test) ? query.test : query.test ? [query.test] : []
-  ).map((test) => get(test))
-
-  const searchTemplates = await getTemplates({ prod: true })
+export const getStaticProps: GetStaticProps<Props> = async () => {
```
jamesgorrie, line 14: This will mean that we don't update except on build, but we can fix that in a perf PR.

rank/pages/index.tsx
```diff
@@ -37,6 +31,13 @@ export const Index: NextPage<Props> = ({ searchTemplates }) => {
           Rank is a tool help ensure the continued quality of our search ranking
           whilst helping us iterate to improve it.
         </p>
+        <p className="mb-2">
```
jamesgorrie, line 34: I'd like to keep this as non-developer as possible if we can?

rank/pages/search.tsx
```diff
@@ -1,59 +1,70 @@
...
+  const indices = await listIndices()
 
-  const data: SearchApiResponse = await fetch(
-    `${origin}/api/search?${reqQs}`
-  ).then((res) => res.json())
+  const query = qs.query ? qs.query.toString() : ''
+  const env = qs.env ? qs.env.toString() : 'local'
+  const index = qs.env ? qs.index.toString() : indices[0]
+
+  let data: SearchResponse = {
```
jamesgorrie, line 31: This looks a little untidy. What are we trying to say here? It looks a little like "If there is not query, there is no data" so why not just make data null?

rank/types/elasticsearch.ts
```diff
@@ -2,106 +2,106 @@
...
 }
 
 export type Hit<DocType = Record<string, any>> = {
-  _id: string
-  _score: string
-  _source: DocType
-  _explanation: Record<string, unknown>
-  highlight: Record<string, string[]>
-  matched_queries: string[]
+    _id: string
```
jamesgorrie, line 89: What's with the weird linting?

rank/services/search-templates.ts
```diff
@@ -93,34 +55,44 @@ export async function getLocalTemplates(): Promise<SearchTemplate[]> {
...
+    const indices = await listIndices()
+    ids = envs.flatMap((env) =>
+      indices.map((index) => `${env}/${index}` as SearchTemplateString)
+    )
+  }
+  const queries = await getQueries()
+  const templates = ids
+    .map((id) => new SearchTemplate(id))
+    .map((template) => {
+      template.query = queries[template.env][template.namespace]
```
jamesgorrie, line 82: Mutating objects like this if often a sign we might want to do it differently.

rank/services/search-templates.ts
```diff
@@ -93,34 +55,44 @@ export async function getLocalTemplates(): Promise<SearchTemplate[]> {
...
   return templates
 }
 
+export async function getTemplate(
+  env: Env,
+  index: Index
+): Promise<SearchTemplate> {
+  const queries = await getQueries()
+  const template = new SearchTemplate(`${env}/${index}`)
+  template.query = queries[template.env][template.namespace]
```
jamesgorrie, line 94: Yea, as we're doing this multiple times, I think there's something up.

jamesgorrie (commented):
It is large, not just in file size, but also what we are changing.

I wonder if it might make more sense to split it into:
* search template simplifications
* script changes
* updating search to service

On top of this - there are a few places where we are mutating objects, which oftenm will complicate things, but I am finding it hard to tell if it's the PR or the specific logic that I am finding fishy.

I've suggested the steps above - but maybe there's a better breakdown?

rank/.prettierrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
   "singleQuote": true,
-  "semi": false
-}
\ No newline at end of file
+  "semi": false,
+  "tabWidth": 2
```
jamesgorrie, line 4: [The default is already 2](https://prettier.io/docs/en/options.html#tab-width)

rank/types/elasticsearch.ts
harrisonpim: not sure tbh - i was having all sorts of linting issues when i started work on this which i managed to resolve locally, but it looks like they've managed to creep back into this PR

rank/.prettierrc.js
harrisonpim: think this was introduced while I was trying to resolve the linting problems i mentioned above ☝️

rank/services/search-templates.ts
harrisonpim: This seemed like the neatest way around the async-class-creation problem we talked about previously. Happy to keep pulling that thread if you're not comfortable with this though!

rank/pages/search.tsx
harrisonpim: totally right!

rank/types/searchTemplate.ts
```diff
@@ -0,0 +1,52 @@
...
+  env: Env
+  query: Query
+  string: SearchTemplateString
+
+  constructor(env: Env, index: Index, query: Query) {
+    this.env = env
+    this.index = index
+    this.query = query
+    this.namespace = getNamespaceFromIndexName(index)
+    this.string = `${this.env}/${this.index}`
```
jamesgorrie, line 29: later: we could put this in the `toString()` method

rank/services/test.ts
```diff
@@ -1,41 +1,35 @@
...
+  const template = await getTemplate(env, index)
   const test = tests[template.namespace].find((test) => test.id === testId)
   if (!test) throw Error(`No such test ${testId}`)
 
   const { cases, metric, searchTemplateAugmentation } = test
 
-  const searchTemplate = searchTemplateAugmentation
-    ? searchTemplateAugmentation(test, { ...template.source })
-    : { ...template.source }
+  template.query = searchTemplateAugmentation
```
jamesgorrie, line 25: Could we leave a note that we shouldn't mutate this here?

rank/types/searchTemplate.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export type ApiSearchTemplateRes = {
+  templates: {
+    id: string
+    index: Index
+    query: string // this is a JSON string
+  }[]
+}
+
+export function getNamespaceFromIndexName(
```
jamesgorrie, line 41: Do we still need this?

jamesgorrie (approved):
Few nitpicks, but all good!

rank/types/searchTemplate.ts
harrisonpim: yeah, it's still used in a few places like the `constructor` above ([see here](https://github.com/wellcomecollection/catalogue-api/pull/267/files/d746e57b8514e25bbf4bed7d792406de1bbf7a17#diff-afe3ec8ad8501ef4d524bbfc4617be0418f3be1e6d50b3f6fd7b70385e097d95R28))

jamesgorrie (approved):

PR #270, opened 2021-09-10
Rename rank's Env type

harrisonpim:
The existence of an `Env` type is a hangover from when we ran tests between `['prod', 'stage']` environments. 

To better reflect the current state of the app, in which we only really test local candidate queries against our remote prod queries (each across multiple indexes), this type could be renamed `QuerySource`, with its options changed to `['production', 'candidate']`


jamesgorrie (commented):
Just checking, we don't mean `queryEnv`? Source is quite overloaded.

harrisonpim:
I think we'd spoken about `querySource` previously, but I'm happy either way - either is better than `Env` imo

jamesgorrie:
I think I would go with `env` as `source` often means the actual content e.g. source of a file.

jamesgorrie (approved):

PR #271, opened 2021-09-10
Rank scripts

harrisonpim:
- Uses `prompts` more widely to constrain the usable indices/configs to the same namespace/sourceindex etc
- Adds a bit more niceness to what's written to the terminal as you work
- Adds a script to monitor the state of a reindex


rank/scripts/createIndex.ts
```diff
@@ -0,0 +1,99 @@
...
+  // Users should be able to create new indices by modifying mappings they've
+  // fetched from the rank cluster, where those source indices still exist.
+  const configFiles = readdirSync('./data/indices/')
+    .filter((fileName) => fileName.includes('.json'))
+    .map((fileName) => parse(fileName).name)
+  const sourceIndices = await listIndices()
+  const validIndices = sourceIndices.filter((value) =>
+    configFiles.includes(value)
+  )
+
```
jamesgorrie, line 22: I might go for naming this similar to how we name queries.

e.g.

```ts
const localIndices = ...
const remoteIdices = ...
const validIndices = ...
```

rank/scripts/createIndex.ts
```diff
@@ -0,0 +1,99 @@
...
+  const sourceIndex = await prompts({
+    type: 'select',
+    name: 'value',
+    message: 'Which index config are you working from?',
+    choices: validIndices.map((choice) => ({ title: choice, value: choice })),
```
jamesgorrie, line 27: Convention is to name the mapped item as to what it is:
```suggestion
    choices: validIndices.map((index) => ({ title: index, value: index })),
```

rank/scripts/createIndex.ts
```diff
@@ -0,0 +1,99 @@
...
+  // fetched from the rank cluster, where those source indices still exist.
+  const configFiles = readdirSync('./data/indices/')
+    .filter((fileName) => fileName.includes('.json'))
+    .map((fileName) => parse(fileName).name)
+  const sourceIndices = await listIndices()
+  const validIndices = sourceIndices.filter((value) =>
+    configFiles.includes(value)
+  )
+
+  const sourceIndex = await prompts({
```
jamesgorrie, line 23: Naming suggestion
```suggestion
  const localIndex = await prompts({
```

rank/scripts/createIndex.ts
```diff
@@ -0,0 +1,99 @@
...
+    message: 'Which index config are you working from?',
+    choices: validIndices.map((choice) => ({ title: choice, value: choice })),
+  }).then(({ value }) => value)
+
+  const indexConfig = await import(`../data/indices/${sourceIndex}.json`).then(
+    (mod) => mod.default
+  )
+
+  const namespace = getNamespaceFromIndexName(sourceIndex)
+  info(`Your new index should have a name like ${namespace}-with-secret-sauce`)
```
jamesgorrie, line 35: I'm not sure what this is saying - maybe
```suggestion
  info(`Your new index should start with it's namespace "works-" or "images-"`)
```

jamesgorrie: As mentioned you could work this out automagically.

rank/scripts/createIndex.ts
```diff
@@ -0,0 +1,99 @@
...
+  const destIndex = await prompts({
```
jamesgorrie, line 36:
```suggestion
  const remoteIndex = await prompts({
```

rank/scripts/getConfig.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import fs from 'fs'
+import { getRankClient } from '../services/elasticsearch'
+import { listIndices } from '../services/search-templates'
+import prompts from 'prompts'
+
+global.fetch = require('node-fetch')
+
+async function go() {
+  const allIndices = await listIndices()
```
jamesgorrie, line 11:
```suggestion
  const remoteIndices = await listIndices()
```

jamesgorrie (dismissed):
Nit picking at some naming conventions and if they can be more similar to other places in the codebase

jamesgorrie (approved):

PR #273, opened 2021-09-13
Rank update tests

harrisonpim:
Uses the refactored types and functionality in the test suite

### DIFF
--- a/rank/jest.setup.js
+++ b/rank/jest.setup.js
@@ -1,0 +1,1 @@
+jest.setTimeout(30000)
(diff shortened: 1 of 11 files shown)


rank/jest.setup.js
```diff
@@ -0,0 +1 @@
+jest.setTimeout(30000)
```
alexwlchan, line 1: It might be worth a comment here explaining how you got to this number, so it's not lost in the commit history. How long would you expect these tests to take? How would I know if I'm hitting timeouts because of gradually getting slower tests or a bug I've just introduced?

harrisonpim: Yep, you're right, will add a comment now!

For context: we know that there are some janky performance problems in rank, though it's unclear whether they're caused by the query/app/cluster. Hoping to dig through them when @jamesgorrie is back next week. 
Jest's default timeout is 5s - too short for any of that weirdness to start manifesting. Bumping that limit to 30s allows the tests to run and opens the door to investigating the performance issues, but we should _definitely_ not be deploying new queries with such a loose threshold.

alexwlchan (approved):

PR #283, opened 2021-09-23
Bump to Elasticsearch v7.14.2

alexwlchan:
This is the latest patch version, which we need to bump so we can create a new cluster with the same version for the 2021-09-23 pipeline.

### DIFF
--- a/terraform/shared/elastic.tf
+++ b/terraform/shared/elastic.tf
@@ -2,7 +2,7 @@ resource "ec_deployment" "catalogue_api" {
   name = "catalogue-api"
 
   region                 = "eu-west-1"
-  version                = "7.14.1"
+  version                = "7.14.2"
   deployment_template_id = "aws-io-optimized-v2"
 
   traffic_filter = local.catalogue_ec_traffic_filter


alicefuzier:
Would it make sense to put this in the same pr with the new index & internal model version? I wouldn't want to merge this if it means we block api deployments until we have an index with that version. We made quite a lot of changes and I don't expect this reindex to go smoothly

alexwlchan:
> Would it make sense to put this in the same pr with the new index & internal model version?

Possibly, although it's already applied.

> I wouldn't want to merge this if it means we block api deployments until we have an index with that version. We made quite a lot of changes and I don't expect this reindex to go smoothly

It doesn't – in fact, uncoupling it from the index/internal model version means we can merge it separately.

Quick background (I think you were away when we hit this bug):

* Elasticsearch is very strict about cross-cluster replication. The target cluster must be on the same-as-or-newer version of Elasticsearch, down to patch versions. If the pipeline is v7.14.2 and the API is v7.14.1, then CCR doesn't work.
* CCR breaks if you upgrade the pipeline cluster. If both pipeline and API are v7.14.1 and then you upgrade the pipeline to v7.14.2, then CCR will silently stop working and updates stop appearing in the API. This happened, we didn't notice, oops.
* So now we export the version of the API cluster via Terraform, so we know the pipeline cluster will always be created with the same version.
* But Elasticsearch is also very strict about which version you can create – you can only create new clusters with the latest patch version of ES. So the API exported v7.14.1, I tried to create a corresponding pipeline cluster, but that's not allowed because v7.14.2 is the current version.

tl;dr: I'm bumping the API cluster to the latest patch version so I can create a new pipeline cluster at the same version, so I can enable CCR between them.

(This is all commented in the pipeline repo, but it's harder to find your way back in the other direction.)

Make sense?

alicefuzier:
Thanks yes, makes sense

alicefuzier (approved):

PR #286, opened 2021-09-29
Return a 5xx if the default Elasticsearch index doesn't exist

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5316

## The context

In the catalogue API, we let users specify an index on their requests by passing the `_index` parameter, e.g.

```
GET /works?_index=my-great-index
```

This is useful for testing changes in indexes that haven't yet gone out, e.g. I think it's how we select the "tei-on" index for the TEI toggle.

It's possible to select an index that doesn't exist with this parameter, so like a well-behaved HTTP API we return a 404 Not Found if we go to query an index and the index doesn't exist.

## The problem

We return a 404 Not Found _even if the user didn't specify an index_. That is, a request to

```
GET /works
```

would return a 404 if the default index is missing. This is wrong – if the default index is missing, that's a problem in the server that we need to fix.

It also means these errors don't get caught by our 5xx alerting. The requests API was serving 404s for almost 24 hours before we noticed, because we'd deleted the underlying Elasticsearch index.

## The solution

We return a 500 Internal Server Error if the index is missing **unless** the user has explicitly told us which index to use.

That means we'll continue to return 404s when somebody asks for a missing index, but if the default index is missing we'll get a 5xx which will be easier to spot in our alerting.

### DIFF
--- a/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchErrorHandler.scala
+++ b/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchErrorHandler.scala
@@ -58,6 +58,22 @@ object ElasticsearchErrorHandler extends Logging {
         serverError("Unknown error", e)
     }
 
+  // If a user specifies an index explicitly, e.g. /works?_index=my-great-index, and
+  // that index doesn't exist, we want to return a 404 error -- they've asked for
+  // something that can't be found.
+  //
+  // If a user doesn't specify an index, e.g. /works, and the default index doesn't exist,
+  // we want to return a 500 error -- something is wrong with our defaults.
+  implicit class RouteOps[T](result: Either[ElasticsearchError, T]) {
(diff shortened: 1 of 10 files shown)


jamesgorrie (dismissed):
I'd be tempted for a `400` error over `404` to show that it's not missing, we are actually not able to process that request.

e.g.
* `/works/not-here` -> 404
* `/works/not-here?index=not-here-either` -> 404

I'd argue we can't tell if the latter is missing, as we can't connect to an index.

Splitting hairs, I'm not sure it matters.

alexwlchan:
> I'd argue we can't tell if the latter is missing, as we can't connect to an index.

We can tell – Elasticsearch returns 404 "index_not_found" exception. We handle it here: https://github.com/wellcomecollection/catalogue-api/blob/b5660c7a15ea43567a446591c307df4061b63282/common/search/src/main/scala/weco/api/search/elasticsearch/ElasticsearchError.scala#L10

PR #287, opened 2021-09-29
Deploy the requests API from the catalogue-api repo

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5316

Currently nothing is deploying the requests API automatically; this fixes that.

### DIFF
--- a/.buildkite/pipeline.deploy-prod.yml
+++ b/.buildkite/pipeline.deploy-prod.yml
@@ -36,6 +36,23 @@ steps:
               "--confirmation-wait-for", 3600 ]
     agents:
       queue: nano
+  - label: "Deploy (requests prod)"
+    branches: "main"
+    plugins:
+      - docker#v3.5.0:
+          image: 760097843905.dkr.ecr.eu-west-1.amazonaws.com/wellcome/weco-deploy:5.6
+          workdir: /repo
+          mount-ssh-agent: true
+          command: [
+              "--project-id", "requests_api",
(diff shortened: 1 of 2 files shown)


jamieparkinson (commented):
How does this relate to https://github.com/wellcomecollection/catalogue-api/pull/253 and https://github.com/wellcomecollection/catalogue-api/pull/257? When did the changes in https://github.com/wellcomecollection/catalogue-api/pull/78 get removed?

alexwlchan:
> When did the changes in #78 get removed?

As part of a commit that apparently went straight to main: https://github.com/wellcomecollection/catalogue-api/commit/6421b6589a99126579af5985de4d2358754616bc#diff-471d1889d5966fa70009332497a3a2fc48d3850163bb6f579f496cb01b1b811c

I think it got pulled out when Robert was trying to lock down the identity account.

Because the requests API deploys into the identity account, the long-term plan was that something else would be deploying the requests API rather than our shared pool of Buildkite instances. Then trying to tighten up their permissions became rather more involved than expected, and we never actually finished that – and somehow putting the requests API back slipped through the cracks.

I discussed this with Alice earlier; the plan is to at least get it deploying again (and we have protection on the buildkite config through the codeowners file now).

> How does this relate to #253 and #257? 

All the weco-deploy config for deploying the requests API is still there, it's just not being exercised by Buildkite.

jamieparkinson (approved):

PR #292, opened 2021-10-08
Rank reindex progress monitor

harrisonpim:
- no longer need to specify a task ID
- get a clearer estimate of the progress of your reindex
- tiny tweaks to the cli utils

```
➜ yarn checkReindex

Still working on reindex from [works-2021-09-27] to [works-candidate][_doc]

 ███████████████░░░░░░░░░░░░░░░░░░░░░░░░░ 37% | ETA: 6116s | 1014000/2695284
```


rank/scripts/checkReindex.ts
```diff
@@ -0,0 +1,42 @@
...
+import cliProgress from 'cli-progress'
+import { getRankClient } from '../services/elasticsearch'
+
+async function go() {
+  const client = getRankClient()
+  const reindexTasks = await client.cat.tasks({
+    actions: 'indices:data/write/reindex',
+  })
+
+  const task_id = reindexTasks.body.split(' ')[1]
```
jamesgorrie, line 12: Could you leave a note here of what we expect as it's a bit opaque.

rank/scripts/checkReindex.ts
```diff
@@ -0,0 +1,42 @@
...
+      {},
+      cliProgress.Presets.shades_classic
+    )
+    progress.start(
+      taskResponse.body.task.status.total,
+      taskResponse.body.task.status.created
+    )
+
+    const timer = setInterval(async function () {
+      taskResponse = await client.tasks.get({ task_id })
```
jamesgorrie, line 29: Not sure of a fix right now, but not sure I like the mutated state of the loop.

PR #300, opened 2021-10-29
Rank on cli

harrisonpim:
Hopefully answers the question 'How do I run rank tests from the command line?' 
ie closes https://github.com/wellcomecollection/platform/issues/5349

PR #301, opened 2021-11-01
Minor improvements to the rank CLI

alexwlchan:
Follows https://github.com/wellcomecollection/catalogue-api/pull/300, a first step towards https://github.com/wellcomecollection/platform/issues/5350

A couple of changes:

* Better error messages if you don't have the appropriate environment variables set
* Allow running the rank tests against the staging index
* Allow choosing where you run the CLI tests. In particular, you can set the `RANK_QUERY_ENVIRONMENT` variable and choose from production/staging/candidate. I was able to see this working correctly by:

    - Messing with my local copy of the candidate queries
    - Setting the env var to `candidate`
    - Watching the tests fail

    I haven't been able to check for a failure against the staging index, but at some point maybe we can deploy a deliberately broken API there and see if the tests catch it.

### DIFF
--- a/rank/__tests__/eval.test.ts
+++ b/rank/__tests__/eval.test.ts
@@ -6,7 +6,7 @@ import {
 import { getTemplates, listIndices } from '../services/search-templates'
 
 import { TestResult } from '../types/test'
-import testService from '../services/test'
+import service from '../services/test'
 import tests from '../data/tests'
 
 global.fetch = require('node-fetch')
@@ -53,7 +53,7 @@ test.each(works)('works.$id', async ({ id }) => {
     (template) => getNamespaceFromIndexName(template.index) === 'works'
   )
 
-  const result = await testService({
+  const result = await service({
     queryEnv: template.queryEnv,
     index: template.index,
     testId: id,
@@ -69,7 +69,7 @@ test.each(images)('images.$id', async ({ id }) => {
     (template) => getNamespaceFromIndexName(template.index) === 'images'
   )
 
-  const result = await testService({
+  const result = await service({
     queryEnv: template.queryEnv,
     index: template.index,
     testId: id,
--- a/rank/__tests__/eval.test.ts
+++ b/rank/__tests__/eval.test.ts
@@ -17,6 +17,8 @@ beforeAll(async () => {
   searchTemplates = await getTemplates()
 })
 
+const queryEnv = process.env.RANK_QUERY_ENVIRONMENT || 'production'
+
 declare global {
   namespace jest {
     interface Matchers<R> {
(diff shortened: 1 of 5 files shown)


alexwlchan:
I don't have access to Vercel yet, so I can't see why deployment has failed. :'(

rank/__tests__/eval.test.ts
```diff
@@ -17,6 +17,8 @@ beforeAll(async () => {
   searchTemplates = await getTemplates()
 })
 
+const queryEnv = process.env.RANK_QUERY_ENVIRONMENT || 'production'
```
harrisonpim, line 20: Not sure what @jamesgorrie's view on this might be, but to me the `queryEnv` and the `index` for testing feel like variables we should set as CLI arguments rather than (as well as? instead of?) environment variables.

alexwlchan: I'd be cool with that (and prefer to do so), I was just too lazy to look up how command-line argument parsing works in Jest. 😛 

Happy to add it if you think the core concept is solid.

jamesgorrie: [yargs is your friend](https://www.npmjs.com/package/yargs)

rank/services/search-templates.ts
```diff
@@ -21,10 +21,12 @@ export async function listIndices(): Promise<Index[]> {
   return indices
 }
 
-export async function getProductionQueries() {
-  const res = await fetch(
-    'https://api.wellcomecollection.org/catalogue/v2/search-templates.json'
-  )
+async function getLiveQueries(environment) {
+  const apiUrl = environment === 'prod'
```
jamesgorrie, line 25:
```suggestion
  const apiUrl = environment === 'production'
```

rank/services/search-templates.ts
```diff
@@ -21,10 +21,12 @@ export async function listIndices(): Promise<Index[]> {
   return indices
 }
 
-export async function getProductionQueries() {
-  const res = await fetch(
-    'https://api.wellcomecollection.org/catalogue/v2/search-templates.json'
-  )
+async function getLiveQueries(environment) {
```
jamesgorrie, line 24: Could we type this argument to `QueryEnv`? This would have caught the bug below.

jamesgorrie: `getEnvironmentQueries`?

rank/__tests__/eval.test.ts
jamesgorrie: Agree with being a arg, probably not worth setting the default and error if it's not set.

rank/services/search-templates.ts
alexwlchan: "caught the bug below"

Which bug?

rank/docs/testing.md
```diff
@@ -4,6 +4,8 @@
 
 To run tests from the CLI, run `yarn test`.
 
+You can choose where to run the rank tests (e.g. staging) by setting the `RANK_QUERY_ENVIRONMENT` environment variable.
```
jamesgorrie, line 7:
```suggestion
```

jamesgorrie (approved):

PR #308, opened 2021-11-11
Make slightly fewer requests when running the rank CLI

alexwlchan:
Previously calling `getTemplate(queryEnv)` would fetch the template for every query environment, then pick out the one you asked for.  This is clearly inefficient – this patch modifies this function to only fetch the template we want, and skip fetching the others.

Locally it takes the CLI from ~44.8s to ~42.3s.

For https://github.com/wellcomecollection/platform/issues/5350. This also shows it might be useful to run rank on pull requests, because you have no idea if the CLI portion actually works.

### DIFF
--- a/.buildkite/scripts/test_rank.sh
+++ b/.buildkite/scripts/test_rank.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash 
+#!/usr/bin/env bash
 
 set -o errexit
 set -o nounset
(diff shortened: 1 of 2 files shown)


harrisonpim (approved):
good point re running the CLI on rank PRs!

alexwlchan:
> good point re running the CLI on rank PRs!

There's a note about it on the main ticket – if you run it against candidate queries that's probably good enough.

PR #310, opened 2021-11-12
Rank test script

harrisonpim:
- small part of https://github.com/wellcomecollection/platform/issues/5350

`yarn test` now has a more coherent set of arguments for being run in CI

Rank can also be run from the CLI with a friendly interface for humans (`yarn rank`), which will prompt the user to supply any arguments which they haven't already. `yarn rank` then calls `yarn test` with those collected args


rank/scripts/createIndex.ts
```diff
@@ -78,20 +72,17 @@ async function go() {
       body: {
         source: {
           index: localIndex,
+          size: 100, // batch size reduced to avoid memory issues
```
alexwlchan, line 75: reduced from what?

rank/scripts/createIndex.ts
```diff
@@ -78,20 +72,17 @@ async function go() {
...
-    if (reindexResp.acknowledged) {
-      success('Reindex started successfully')
-      info('You can monitor the reindex task by running:')
-      code('yarn checkReindex')
-    } else {
-      error('Failed to start reindex with error:')
-      console.info(reindexResp.error)
-    }
+
+    success('Reindex started successfully')
```
alexwlchan, line 83: Why have you removed the error handling code?

rank/scripts/rank.ts
```diff
@@ -0,0 +1,49 @@
...
+import { tests as possibleTests } from '../data/tests'
+
+global.fetch = require('node-fetch')
+
+async function go() {
+  const indices: Index[] = await listIndices()
+
+  const possibleTestIds = Object.values(possibleTests)
+    .flatMap((testSet) => testSet.map((test) => test.id))
+    .filter((v, i, a) => a.indexOf(v) == i) // keep unique
```
alexwlchan, line 15: What are these values? Why does filtering them keep them unique?

alexwlchan (approved):

rank/scripts/createIndex.ts
harrisonpim: longer comment added in https://github.com/wellcomecollection/catalogue-api/pull/310/commits/e6abd8e59aaa92d686c2afa1783fe4b6faebbe4d !

rank/scripts/rank.ts
harrisonpim: longer comment added in https://github.com/wellcomecollection/catalogue-api/pull/310/commits/e6abd8e59aaa92d686c2afa1783fe4b6faebbe4d !

rank/scripts/createIndex.ts
harrisonpim: The error handling just wasn't working here - starting a reindex successfully would throw errors, etc.
I'm planning to revisit these scripts next week to make wider use of the `gatherArgs()` function and tighten up error handling across the board. Won't be gone for long!

PR #312, opened 2021-11-17
query the search.identifiers field

harrisonpim:
- the corresponding query change for https://github.com/wellcomecollection/catalogue-pipeline/pull/1974


harrisonpim:
closed as a result of discussion [here](https://github.com/wellcomecollection/platform/issues/5368)

PR #324, opened 2021-12-08
Provide a better error if a user's account has expired and they request an item

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7451


jamieparkinson (approved):
Excellent!

PR #352, opened 2022-02-10
Have the APIs read from the pipeline cluster, not from their config

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5430

This modifies the Scala applications to fetch their config from values in Secrets Manager (see https://github.com/wellcomecollection/catalogue-pipeline/pull/2035; I've added them by hand for now) rather than from their environment config.

I still need to twiddle some IAM permissions, but this should drop into the existing APIs – and blue/green deployment will protect us if I've done it wrong.


alexwlchan:
Closing as superseded.

PR #359, opened 2022-02-24
Filter partof by title

paul-butcher:
Part of https://github.com/wellcomecollection/platform/issues/5408

This introduces the ability to query partOf.title

### DIFF
--- a/search/src/main/scala/weco/api/search/services/WorksRequestBuilder.scala
+++ b/search/src/main/scala/weco/api/search/services/WorksRequestBuilder.scala
@@ -201,8 +201,14 @@ object WorksRequestBuilder
              If this does cause unwanted results in "real life", then we can consider storing a
              separate non-analysed version of title for term matching.
              */
-            matchPhraseQuery(field = "state.relations.ancestors.title", value = search_term),
-            termQuery(field = "state.relations.ancestors.id", value = search_term)
+            matchPhraseQuery(
+              field = "state.relations.ancestors.title",
+              value = search_term
+            ),
+            termQuery(
+              field = "state.relations.ancestors.id",
+              value = search_term
+            )
           ),
           Seq.empty
         )
--- a/search/src/main/scala/weco/api/search/services/WorksRequestBuilder.scala
+++ b/search/src/main/scala/weco/api/search/services/WorksRequestBuilder.scala
@@ -187,7 +187,7 @@ object WorksRequestBuilder
          rather than have to introduce a new filter to the API to cover it.
          */
         bool(
-          Seq(),
+          Nil,
           Seq(
             termQuery(
               field = "state.relations.ancestors.title.keyword",
@@ -198,7 +198,7 @@ object WorksRequestBuilder
               value = search_term
             )
           ),
-          Seq()
+          Nil
         )
 
       case AvailabilitiesFilter(availabilityIds) =>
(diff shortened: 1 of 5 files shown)


search/src/main/scala/weco/api/search/services/WorksRequestBuilder.scala
```diff
@@ -179,8 +179,28 @@ object WorksRequestBuilder
           field = "data.items.locations.locationType.id",
           values = itemLocationTypeIds
         )
-      case PartOfFilter(id) =>
-        termQuery(field = "state.relations.ancestors.id", value = id)
+      case PartOfFilter(search_term) =>
```
paul-butcher, line 182: I think this whole match block is ripe for refactoring, but I won't do it now as part of this change.

It basically amounts to a chain of `if object is of class X, do this else if object is of class Y do that`.

I think it would be better for each of these cases to be a method

jamieparkinson: I would be slightly hesitant about that sort of change; we used to have a very DRY and modular class for building queries, and nobody - especially especially non-Scala devs - could reason about how it related to the JSON that elastic4s produces. 

The advantage of doing it like this is that it looks pretty much like the JSON that it produces, and it's easier for eg @harrisonpim to make changes. I think we're gravitating towards the view that the API shouldn't be written in Scala for exactly this sort of reason, and that some sort of language that is better at handling JavaScript Object Notation might be more appropriate and work better for the team...

jamieparkinson: I would be slightly hesitant about that sort of change; we used to have a very DRY and modular class for building queries, and nobody - especially especially non-Scala devs - could reason about how it related to the JSON that elastic4s produces. 

The advantage of doing it like this is that it looks pretty much like the JSON that it produces, and it's easier for eg @harrisonpim to make changes. I think we're gravitating towards the view that the API shouldn't be written in Scala for exactly this sort of reason, and that some sort of language that is better at handling JavaScript Object Notation might be more appropriate and work better for the team...

search/src/main/scala/weco/api/search/services/WorksRequestBuilder.scala
```diff
@@ -179,8 +179,28 @@ object WorksRequestBuilder
...
+        /*
+         The partOf filter matches on either the id or the title of any ancestor.
+         As it is vanishingly unlikely that an id_minter canonical id will coincidentally
+         match the title of a Work or Series, this query can be a simple OR (i.e. bool.should),
+         rather than have to introduce a new filter to the API to cover it.
+         */
+        bool(
+          Seq(),
```
jamieparkinson, line 190: Scala bikeshed: consider `Nil`

jamieparkinson (approved):

search/src/main/scala/weco/api/search/services/WorksRequestBuilder.scala
paul-butcher: Fair enough.  I guessed that there would be a reason behind it.  I just find this pattern a bit odd

Shifting to another language would also help to remove the internal_model dependency.  If we can make it so it returns data (almost) verbatim from the database, then may not even need types.

PR #364, opened 2022-03-10
Expand hold request error messaging to allow for self registered patronType (29)

melanierogan:
This PR takes the `lookupPatronType` method from `scala libs` and uses it to allow for cases where we have self registered users (type 29). Self registered users, who can't request items, will now get a more readable and helpful error message when they do try to request an item. 

This should cover the last of the work for https://github.com/wellcomecollection/wellcomecollection.org/issues/7645


requests/src/main/scala/weco/api/requests/services/SierraRequestsService.scala
```diff
@@ -182,6 +194,17 @@ class SierraRequestsService(
...
+  private def checkIfUserIsSelfRegistered(
+    patron: SierraPatronNumber
+  ): Future[Option[HoldRejected]] =
+    sierraSource
+      .lookupPatronType(patron)
+      .map {
+        case Right(Some(int)) if int == 29 =>
+          Some(HoldRejected.UserIsSelfRegistered)
+        case _ => None
+      }
```
jamieparkinson, line 206: I think it might be better to remove the "opinion" of the type of `HoldRejected` from this method and make it - as its naming kind of encourages - return a `Future[Boolean]`, something like:

```scala
  private def checkIfUserIsSelfRegistered(
    patron: SierraPatronNumber
  ): Future[Boolean] =
    sierraSource
      .lookupPatronType(patron)
      .map {
        case Right(Some(29)) => true
        case _               => false
      }
```
in this case maybe the naming could be simplified, eg to `userIsSelfRegistered`

requests/src/main/scala/weco/api/requests/services/SierraRequestsService.scala
```diff
@@ -71,6 +75,14 @@ class SierraRequestsService(
       //      so we look to see if this item can be requested.
       //    - The user's patron record has expired and they can no longer make requests.
       //
+      case Left(SierraErrorCode(132, 2, 500, _, Some(description)))
+          if description.contains("You may not make requests") =>
+        checkIfUserIsSelfRegistered(patron)
+          .map {
+            case Some(rejected) => Left(rejected)
+            case None           => Left(HoldRejected.UserIsSelfRegistered)
+          }
```
jamieparkinson, line 84: I think(?) this in combination with `checkIfUserIsSelfRegistered` as implemented below will always return a `HoldRejected.UserIsSelfRegistered`, which isn't what we want - it should be an `UnknownReason` if this error occurs for any patron other than a self-reg patron. I don't believe we'll see users with no pType, but we definitely want to know about it  (in the form of a 500 error) if we do, and anyway the error message above only really makes sense for self-reg users.

That is to say I think the behaviour should be: if we get a 132 error which says 'You may not make requests' _and_ if the patron is self-registered, we should return a `HoldRejected`. Otherwise, we don't know what to say to the user, so we should return an `UnknownReason`.

jamieparkinson (changes requested):
This is a great first Scala PR for a new dev to a large and slightly peculiar codebase! Just some comments about desired behaviour and perhaps a way of implementing it that separates concerns a little more.

jamieparkinson (approved):
✨

PR #370, opened 2022-03-31
bump elastic index and model

paul-butcher:
for today's reindex


alexwlchan (approved):

common/display/src/main/scala/weco/catalogue/display_model/ElasticConfig.scala
```diff
@@ -35,7 +35,7 @@ trait ElasticConfigBase {
   //
   // The different config allows applications to choose whether they want to read
   // from the pipeline cluster or the API cluster.
-  val pipelineDate = "2022-03-18"
+  val pipelineDate = "2022-03-31"
```
alexwlchan, line 38: Do you need to set an `a` suffix if it's `2022-03-31a`?

PR #374, opened 2022-04-06
Set up an API that doesn't need CCR

alexwlchan:
The second half of https://github.com/wellcomecollection/platform/issues/5430

This gets the pieces in place for a pipeline+API that doesn't use CCR. In particular:

* The API will now select a set of secrets to use at startup and use the correct set of secrets based on `ElasticConfig`, rather than having them hard-coded in Terraform

### DIFF
--- a/common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
+++ b/common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
@@ -48,8 +48,8 @@ object PipelineElasticClientBuilder {
   }
 
   private def getSecretString(
-                               id: String
-                             )(implicit secretsClient: SecretsManagerClient) = {
+    id: String
+  )(implicit secretsClient: SecretsManagerClient) = {
     val request =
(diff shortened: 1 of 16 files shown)


common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
```diff
@@ -1,22 +1,22 @@
...
 import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRequest
 import weco.catalogue.display_model.PipelineClusterElasticConfig
 import weco.elasticsearch.ElasticClientBuilder
 
-object SnapshotGeneratorElasticClientBuilder {
+object PipelineElasticClientBuilder {
 
   // We create a new pipeline cluster for every pipeline, and we want the
-  // snapshot generator to read from that cluster.
+  // to read from that cluster.
```
paul-butcher, line 12: we want the ... to read

paul-butcher (approved):
LGTM, just one word missing from a comment.

PR #375, opened 2022-04-07
Don't use Elasticsearch to look up works in the items API

alexwlchan:

This decouples the items API from Elasticsearch. Previously it went:

```mermaid
graph LR
    I[items API] --> ES[Elasticsearch]
```

Now it goes:

```mermaid
graph LR
    I[items API] --> C[catalogue API]
    C --> ES[Elasticsearch]
```

This means changes to the Elasticsearch infrastructure don't affect the items API. The Scala types aren't quite ideal for this, but they're good enough, and at some point we might revisit this more substantially as part of Catalogue API V3. We could do a more extensive refactor here to entirely decouple it from the Catalogue API, but I don't want to bite off too much in one go.

The diff isn't as big as it appears at first blush – a lot of it is de-denting tests where we no longer need to set up as much context.


alexwlchan:
Oof, building the items API took ~3m in a previous build (https://buildkite.com/wellcomecollection/catalogue-api/builds/1533), now it's done to ~2m. A whole minute just to pull an Elasticsearch container! ⚡

common/stacks/src/main/scala/weco/api/stacks/models/SierraItemIdentifier.scala
```diff
@@ -28,7 +29,7 @@ object SierraItemIdentifier {
       ontologyType = "Item"
     )
 
-  def fromSourceIdentifier(
+  def fromOldSourceIdentifier(
```
jamieparkinson, line 32: what does `old` mean?

items/src/main/scala/weco/api/items/services/WorkLookup.scala
```diff
@@ -1,30 +1,68 @@
...
+        case StatusCodes.NotFound =>
+          info(s"Not Found for GET to $path with $params")
+          Future(Left(WorkNotFoundError(id)))
+
+        case StatusCodes.Gone =>
+          info(s"Gone for GET to $path with $params")
+          Future(Left(WorkGoneError(id)))
+
+        case status =>
+          val err = new Throwable(s"$status from bag tracker API")
```
jamieparkinson, line 57: assuming that `bag tracker` is a bad copy/paste from the storage service?

jamieparkinson (approved):
Remarkably straightforward to do this - nice

common/stacks/src/main/scala/weco/api/stacks/models/SierraItemIdentifier.scala
alexwlchan: "old" is just "the old way of doing things", a change I made to keep the requests API compiling – I didn't want to change too much in one go. The next PR will remove this method when I flip the requests API to use the new types.

items/src/main/scala/weco/api/items/services/WorkLookup.scala
alexwlchan: Oops, yes. Good spot, thanks!

PR #377, opened 2022-04-07
Use the catalogue API to look up items in the requests API

alexwlchan:

I think this works, but it shows a certain amount of creakiness – the data model is a bit wonky. Looking up the work data by querying the item identifier is a bit weird; if we could record the work ID as part of the hold this would get a lot simpler (although that has other issues – we'd still need this code as a fallback path).

There needs to be a Terraform change to the requests API, so it gets the catalogue API URL as an environment variable, but I think the API side is working.


jamieparkinson (approved):
re your comments on work IDs, annoyingly holds can only take one `note`, which needs to be human-readable - so it'll be a pain if we want to use it for that as well as pickup date

alexwlchan:
> re your comments on work IDs, annoyingly holds can only take one note, which needs to be human-readable - so it'll be a pain if we want to use it for that as well as pickup date

Yeah, and we need to accommodate holds made outside our system, so it's probably a non-starter, alas.

PR #378, opened 2022-04-07
introduce partOf.title filter

paul-butcher:
Introduce an explicit partOf.label filter (instead of just partOf).

Once this is deployed, I can change wellcomecollection.org to use it, then remove the joint id/label implementation of partOf


jamieparkinson (commented):
How come this isn't `partOf.title`? ie, matching the data shape (afaict)

jamieparkinson:
We had some issues with inconsistencies between the data model, the filter names, and the aggregation names a while ago, and we came up with [this RFC](https://github.com/wellcomecollection/docs/tree/main/rfcs/037-api-faceting-principles#principles) for a "formula" for how we name them

jtweed:
Sorry, @jamieparkinson is completely correct. It should be `partOf.title`, my mistake from mixing work properties up with subjects and genres...

jamieparkinson (approved):

PR #379, opened 2022-04-08
Query the per-pipeline clusters in the API diff tool

alexwlchan:

This allows the catalogue API diff tool to query the per-pipeline clusters, rather than going to the API cluster. This is the current failure in the "deploy prod" task – the diff tool is looking in the API cluster for the 2022-04-04 index, which doesn't exist there.

### DIFF
--- a/diff_tool/diff_tool.py
+++ b/diff_tool/diff_tool.py
@@ -103,7 +103,10 @@ class ApiDiffer:
         try:
             return (response.status_code, response.json())
         except json.JSONDecodeError:
-            print(f"Non-JSON response received from {url}:\n---\n{response.text}\n---\n", file=sys.stderr)
+            print(
+                f"Non-JSON response received from {url}:\n---\n{response.text}\n---\n",
+                file=sys.stderr,
+            )
             sys.exit(1)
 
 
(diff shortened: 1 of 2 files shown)


diff_tool/diff_tool.py
```diff
@@ -148,7 +153,7 @@ def _display_in_console(stats, diffs):
 @click.option("--console", is_flag=True, help="Print results in console")
 def main(routes_file, console):
     session = api_stats.get_session_with_role(
-        role_arn="arn:aws:iam::756629837203:role/catalogue-ci"
+        role_arn="arn:aws:iam::760097843905:role/platform-ci"
```
jamieparkinson, line 159: This loss of separation is a bit of a shame!

jamieparkinson (approved):

diff_tool/diff_tool.py
alexwlchan: Yeah, it'd be nice to get everything in the catalogue account. One day…

In the meantime, I'm tempted to go for a much broader separation here. Having the diff tool know where where the API is looking is a bit weird – it's odd that it's coupled to the implementation of the API, even though the whole point of it is that it's testing the public-facing behaviour!

I'm tempted to add an undocumented API endpoint `/_workTypes` that returns the aggregation we want here, so we know we're always looking at the same index as the API – and then the diff tool doesn't need to know any secrets, because it can get everything it needs by querying HTTP endpoints.

PR #388, opened 2022-04-22
Enable search API to run locally

paul-butcher:
This change is a first step towards running the API application locally and against a local database.

To come: 
* Instructions
* Simple setup (e.g. Docker ES, configure index and load data)
* The other API applications


common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
```diff
@@ -6,15 +6,57 @@ import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueReques
...
+  * Build an ElasticClient that will connect to the appropriate database
+  * as specified for this environment.
+  *
+  * The settings may come either from environment variables
+  * (if API_ELASTIC_SETTINGS_FROM_ENVIRONMENT is true)
+  * or from the AWS Secrets Manager
+  */
 object PipelineElasticClientBuilder {
+  def apply(serviceName: String): ElasticClient =
+    sys.env.get("API_ELASTIC_SETTINGS_FROM_ENVIRONMENT") match {
```
alexwlchan, line 21: As discussed, let's change this to something like `API_SETTINGS_MODE` which takes value `environment|secretsmanager`, and fails to start if it gets an unrecognised value.

common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
```diff
@@ -6,15 +6,59 @@ import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueReques
...
+  * as specified for this environment.
+  *
+  * The settings may come either from environment variables
+  * (if API_SETTINGS_MODE is environment) or from the AWS Secrets Manager
+  */
 object PipelineElasticClientBuilder {
+  def apply(serviceName: String): ElasticClient =
+    sys.env.get("API_SETTINGS_MODE") match {
+      case None | Some("secretsmanager") =>
+        SecretsElasticClientBuilder(serviceName)
```
alexwlchan, line 20: This is fine for now, but at some point I think we should update the API Terraform to pass this setting, so we can remove the default and always be explicit.

common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
```diff
@@ -6,15 +6,59 @@ import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueReques
...
+    }
+}
+
+/**
+  * Build an ElasticClient from Environment Variables
+  * Use this to run the API without accessing the AWS Secrets Manager.
+  * For example, to run against a non-pipeline database such as a local copy.
+  */
+object EnvElasticClientBuilder {
+  def apply(serviceName: String): ElasticClient = {
```
alexwlchan, line 36: Do we need to pass the `serviceName` variable here?

common/search/src/main/scala/weco/api/search/config/builders/MetricsBuilder.scala
```diff
@@ -0,0 +1,31 @@
...
+  * For a "Real" instance, publish metrics on CloudWatch,
+  * For a local or experimental instance, simply store them in memory.
+  */
+object MetricsBuilder {
+  def apply(config: Config)(
+    implicit
+    materializer: Materializer,
+    ec: ExecutionContext
+  ): Metrics[Future] =
+    sys.env.get("API_USE_MEMORY_METRICS") match {
```
alexwlchan, line 25: As with API_SETTINGS_MODE, I think we'd be better off changing this to a MODE setting rather than a boolean.

In particular, I can imagine it might be useful to change this to a third implementation of Metrics that logs to stdout (I think MemoryMetrics holds the values but doesn't send them anywhere).

alexwlchan (commented):
I think this could be merged as-is, but some suggestions that could be addressed here or in a future PR.

common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
jamieparkinson: Is it necessary to specify where the secrets come from? A lot of tools just use a predefined precedence (ie here we'd get from environment or, if that wasn't present, secrets manager)

PR #389, opened 2022-04-22
Move all the include-handling logic out of DisplayWork

alexwlchan:
Another step towards https://github.com/wellcomecollection/platform/issues/5449; follows #387

Previously we added the CatalogueJsonUtil trait, which was a thin wrapper around:

```scala
for {
  displayWork <- DisplayWork(internalWork, includes)
  resultJson <- displayWork.toJson
}
```

This patch strips a lot of the include-handling logic out of the display library (not all of it; I haven't sorted images yet), so we get something more like:

```scala
for {
  displayWork <- DisplayWork(internalWork)  // no includes!
  fullJson    <- displayWork.toJson
  resultJson  <- fullJson.removeKeysBasedOn(includes)
}
```

The interesting parts of this patch are:

- `JsonOps`, which has the Circe code for manipulating JSON, and associated tests
- `CatalogueJsonUtil`, which calls `JsonOps` to remove keys from the DisplayWork JSON based on the includes. It also inherits a lot of the tests that were previously in `DisplayWorkSerialisationTest`
- `DisplayWork`, which has had a lot of include-related logic stripped out

Although it looks large, a lot of it is in the new tests – it's hard to test you're manipulating JSON correctly without dropping in large blocks of JSON. 😅 

The next step is to repeat all this for images, but this is quite enough for one patch.


common/display/src/main/scala/weco/catalogue/display_model/models/GetIdentifiers.scala
```diff
@@ -3,10 +3,15 @@ package weco.catalogue.display_model.models
...
 
 trait GetIdentifiers {
-  protected def getIdentifiers(id: IdState, includesIdentifiers: Boolean) =
+  protected def getIdentifiers(id: IdState): List[DisplayIdentifier] =
+    id.allSourceIdentifiers.map(DisplayIdentifier(_))
+
+  protected def getIdentifiers(
+    id: IdState,
+    includesIdentifiers: Boolean
+  ): Option[List[DisplayIdentifier]] =
```
paul-butcher, line 12: Given the changes in DisplayWork, which I assume mean that if those list values were absent, they are now empty lists, would it be appropriate for this to behave in the same way?

i.e.  not `Option[List...] ... else None`, but always return a lis, which will be empty if includesIdentifiers is false.

Is there a meaningful difference elsewhere between the identifiers option/list being None vs being empty?

(even if this would be better as an empty list, I can see that it would be a bit too widespread a change to include in this pull)

alexwlchan: Unfortunately not, or at least not without changing the outward behaviour of the API.

If you fetch images `/images?include=contributors`, it includes the contributors from the source work. Contributors can be identified, but there's no way to tell the images endpoint "please include/exclude identifiers in the response", so we omit identifiers on images.

When we construct a DisplayContributor from an internal model Contributor, we pass `includeIdentifiers = true|false` depending on whether we're in a Work or an Image context.

jamieparkinson (approved):

PR #390, opened 2022-04-25
Always serialise image responses through CatalogueJsonUtil

alexwlchan:
Another step towards https://github.com/wellcomecollection/platform/issues/5449; follows https://github.com/wellcomecollection/catalogue-api/pull/389

The images API used to pass includes into the display model, then serialise it to JSON:

```scala
for {
  displayImage <- DisplayImage(internalImage, includes)
  resultJson   <- displayImage.toJson
}
```

This patch follows a pattern now used in the works API – always construct a full display model, then use the includes to add/remove keys:

```scala
for {
  displayImage <- DisplayImage(internalImage, includes = ImageIncludes.all)
  fullJson     <- displayImage.toJson
  resultJson   <- fullJson.updateKeysBasedOn(includes)
}
```

This is a little bit different to the works API, because we have to deal with similar images on the single image endpoint – we're adding values as well as removing them. I'll highlight where that happens in the diff.


search/src/main/scala/weco/api/search/json/CatalogueJsonUtil.scala
```diff
@@ -18,33 +13,91 @@ trait CatalogueJsonUtil {
...
+                      images =>
+                        images.map(DisplayImage(_, SingleImageIncludes.all))
+                    )
+                    .asJson
+                )
+              case None => jsonObj
+            }
+        )
+      else
+        j
```
alexwlchan, line 96: When the images index contains display images, we can modify this method to take `value: Option[Seq[Json]]`, where the JSON objects are whatever we read out of the `"display"` field in the index.

jamieparkinson (approved):

PR #406, opened 2022-05-05
Convert WorksServiceTest to use JSON test documents

alexwlchan:

I've added the JSON test documents to this repo already; this converts WorksServiceTest to start using them. Although it's still indexing against the old Elasticsearch structure, when we're ready to flip to the new index structure, this test should be relatively unchanged.

This, hopefully, is sufficiently small as to be reviewable, and look how much smaller it is! 🔥

### DIFF
--- a/search/src/test/scala/weco/api/search/fixtures/TestDocumentFixtures.scala
+++ b/search/src/test/scala/weco/api/search/fixtures/TestDocumentFixtures.scala
@@ -1,0 +1,82 @@
+package weco.api.search.fixtures
+
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s.Index
+import io.circe.Json
+import org.scalatest.Suite
+import org.scalatest.concurrent.PatienceConfiguration.Timeout
+import org.scalatest.time.{Seconds, Span}
+import weco.catalogue.internal_model.work.{Work, WorkState}
+import weco.catalogue.internal_model.Implicits._
+import weco.elasticsearch.test.fixtures.ElasticsearchFixtures
+import weco.fixtures.LocalResources
+import weco.json.JsonUtil._
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.{Failure, Success, Try}
+
+trait TestDocumentFixtures extends ElasticsearchFixtures with LocalResources {
+  this: Suite =>
+
+  val visibleWorks = (0 to 4).map(i => s"works.visible.$i")
+  val invisibleWorks = (0 to 2).map(i => s"works.invisible.$i")
+  val redirectedWorks = (0 to 1).map(i => s"works.redirected.$i")
+  val deletedWorks = (0 to 3).map(i => s"works.deleted.$i")
+
+  val works = visibleWorks ++ invisibleWorks ++ redirectedWorks ++ deletedWorks
+
+  val worksFormatBooks = (0 to 3).map(i => s"works.formats.$i.Books")
+  val worksFormatJournals = (4 to 6).map(i => s"works.formats.$i.Journals")
+  val worksFormatAudio = (7 to 8).map(i => s"works.formats.$i.Audio")
+  val worksFormatPictures = (9 to 9).map(i => s"works.formats.$i.Pictures")
+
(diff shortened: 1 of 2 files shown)


search/src/test/scala/weco/api/search/services/WorksServiceTest.scala
```diff
@@ -82,52 +58,29 @@ class WorksServiceTest
...
-        allWorks = indexedWorks(count = 3),
+        allWorks = visibleWorks,
         expectedWorks = Seq(),
-        expectedTotalResults = 3,
+        expectedTotalResults = visibleWorks.length,
         worksSearchOptions = createWorksSearchOptionsWith(pageNumber = 4)
       )
     }
 
-    it("does not list invisible works") {
```
alexwlchan, line 92: I removed this test because it's now covered by "gets records in Elasticsearch" – that earlier test indexes a mix of visible/invisible/redirected/deleted works, and checks that only the visible ones come back.

search/src/test/scala/weco/api/search/services/WorksServiceTest.scala
```diff
@@ -149,242 +102,99 @@ class WorksServiceTest
...
+        expectedWorks = List("work-title-dodo"),
         expectedTotalResults = 1,
         worksSearchOptions = createWorksSearchOptionsWith(
-          searchQuery = Some(SearchQuery("emu \""))
+          searchQuery = Some(SearchQuery("dodo \""))
         )
       )
     }
 
-    it("returns results in consistent sort order") {
```
alexwlchan, line 186: I binned this test because it's a bit of a faff in the new structure, and I don't think we care about it that much. It's also tested elsewhere.

search/src/test/scala/weco/api/search/services/WorksServiceTest.scala
```diff
@@ -149,242 +102,99 @@ class WorksServiceTest
...
+        expectedTotalResults = visibleWorks.size,
         worksSearchOptions = createWorksSearchOptionsWith(
-          filters = List(FormatFilter(List(ManuscriptsAsian.id, Books.id)))
+          pageSize = 2,
+          pageNumber = 2
         )
       )
     }
 
-    it("filters results by item locationType") {
```
alexwlchan, line 303: I removed these tests because they're not specific to WorksService.

search/src/test/scala/weco/api/search/fixtures/TestDocumentFixtures.scala
```diff
@@ -0,0 +1,88 @@
...
+  ): Unit = {
+    val documents = getTestDocuments(documentIds)
+
+    val result = elasticClient.execute(
+      bulk(
+        documents.map { fixture =>
+          indexInto(index.name)
+            .id(fixture.id)
+            .doc(toJson(fixture.work).get)
+//            .doc(fixture.document.noSpaces)
```
paul-butcher, line 68: Is this commented line dead, or just resting?

alexwlchan: Just resting – this is what the code in this fixture will eventually look like, because this line tells ES to index the new document structure.

search/src/test/scala/weco/api/search/services/WorksServiceTest.scala
```diff
@@ -413,49 +221,34 @@ class WorksServiceTest
...
-              List(
-                AggregationBucket(data = Books, count = 2),
-                AggregationBucket(data = ArchivesAndManuscripts, count = 1),
-                AggregationBucket(data = Audio, count = 1)
-              )
+      val expectedAggregations = WorkAggregations(
+        Some(
+          Aggregation(
+            List(
+              AggregationBucket(data = Format.Books, count = 4),
```
paul-butcher, line 228: Should this be `count=worksFormatBooks.length` etc.

paul-butcher (approved):
Just one minor thing about magic numbers (which you can take or leave as you see fit), but otherwise good to go

search/src/test/scala/weco/api/search/services/WorksServiceTest.scala
alexwlchan: Ooh, yes, good suggestion.

PR #411, opened 2022-05-06
Update WorksIncludesTest to use raw JSON responses

alexwlchan:

This test gets a lot bigger, but I think it's also clearer and simpler. Rather than having clever JSON helpers, we just have hard-coded responses that we expect to see. This makes it easier to see what the test is asserting because there's no obfuscation or helpers in the way, and reduces a risk of a bug in the helpers masking an actual API bug.

As a bonus, if we port these tests to TypeScript it'll just be a copy/paste job.


alexwlchan:
Hmm! The tests locally and in CI are returning notes in a different order, I wonder why.

PR #417, opened 2022-05-09
Use hard-coded JSON responses for some more includes tests

alexwlchan:

### DIFF
--- a/search/src/test/scala/weco/api/search/works/WorksIncludesTest.scala
+++ b/search/src/test/scala/weco/api/search/works/WorksIncludesTest.scala
@@ -824,45 +824,127 @@ class WorksIncludesTest
     ) {
       withWorksApi {
         case (worksIndex, routes) =>
-          val contributors1 =
-            List(Contributor(Person("Ginger Rogers"), roles = Nil))
-          val contributors2 =
-            List(Contributor(Person("Fred Astair"), roles = Nil))
-          val work1 =
-            indexedWork(canonicalId = canonicalId1).contributors(contributors1)
-          val work2 =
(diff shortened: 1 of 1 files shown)


search/src/test/scala/weco/api/search/works/WorksIncludesTest.scala
```diff
@@ -571,41 +563,181 @@ class WorksIncludesTest
...
-            )}],
-                   "genres": [ ${genres(genres2)}]
-                 }
-                ]
-              }
-            """
+          assertJsonResponse(routes, path = s"$rootPath/works?include=genres") {
+            Status.OK ->
+              s"""
+                 |{
```
paul-butcher, line 571: IIUIC, both JSONs are parsed by assertJsonResponse and compared as objects, so preserving the "correct" indentation with `|` isn't needed.

Having the `|` prefix and stripMargin at the end makes it look like assertJsonResponse is directly comparing JSON strings, giving the impression that these tests could be a bit fragile (e.g. that it would break if we change the indentation in the API responses).

alexwlchan: 👍

I think that's auto-added by IntelliJ so I didn't really think about it; removing it is sensible so I'll do that.

alexwlchan: Oh, and yes, you've understood correctly.

alexwlchan: I think this is sorted now, good to merge?

jamieparkinson (approved):

PR #428, opened 2022-05-10
Actually use the "display" field to serialise works API responses

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5449; follows https://github.com/wellcomecollection/catalogue-api/pull/427

Now we've ported all the tests, this is a remarkably small change that seems to work almost immediately.

I still haven't done the images side, but this will point the staging API at a new index reading the "display" field for works. I'll deploy it to prod ASAP (if it passes the diff tool), so we can find out if anything has gone wrong.


jamieparkinson (approved):
This is cool

PR #430, opened 2022-05-12
Port ImagesSimilarityTest to the new style of tests

alexwlchan:

This only touches a single Scala test; the rest is entirely the JSON test documents.

I've also started moving the expected responses into JSON files, because they're getting increasingly large. It's easier to read them in a proper JSON editor than when they're embedded as strings in a Scala file.


paul-butcher (approved):
This looks like it exactly replicates what was there before, so this isn't a request for a change.  Maybe something for another ticket.

It looks like each of these tests loads six images, requests one of them, and expects to see the remaining five in the visuallySimilar/withSimilarFeatures/withSimilarColors list.

This is fine as a positive test, showing that the lists exist and are populated on request, but does not demonstrate any discrimination - it does not exclude any dissimilar images.

All of them would pass if the API simply returned all the images in the database in whichever list was requested in the query string (or something more subtle).

Perhaps the tests could be crafted so that the same input data is used for all three, grouped thus:

1. features a colours x
2. features a colours y
3. features b colours y

Then visually similar could return group 1, similar features groups 1 and 2, similar colours groups 2 and 3.

PR #433, opened 2022-05-12
Get the filter-related tests passing in the new style

alexwlchan:

There are a few cases where the tests aren't quite as complete as they could be, for reasons described in #430. For now I've left a TODO comment and I'll come back to them later; this refactor is nearly done and I want to finish it.


paul-butcher (commented):
Should there be something in stacks/README.md?

alexwlchan:
> Should there be something in stacks/README.md?

No, or at least I wasn't planning to add anything; I must have checked it in my accident. Removed.

search/src/test/scala/weco/api/search/images/ApiImagesTestBase.scala
```diff
@@ -1,14 +1,19 @@
...
+import weco.api.search.models.request.SingleImageIncludes
 import weco.catalogue.display_model.test.util.DisplaySerialisationTestBase
 import weco.catalogue.internal_model.generators.ImageGenerators
 import weco.catalogue.internal_model.image._
 
 trait ApiImagesTestBase
     extends ApiTestBase
     with DisplaySerialisationTestBase
-    with ImageGenerators {
+    with ImageGenerators
```
paul-butcher, line 14: Is ImageGenerators still needed? I think you've deleted all the lines that use it.

alexwlchan: Not in this trait, but it is used in other tests that extend from this one, and removing it will cause conflicts in other PRs – I'm going to leave it for now, and it'll come out when I fully strip out internal model.

(The way we handle traits in tests is a bit messy.)

paul-butcher (approved):

PR #436, opened 2022-05-12
Don't use internal work for CatalogueJsonUtilTest

alexwlchan:
This object is manipulating vanilla JSON objects; there's nothing work-specific about it.  By operating directly on blobs of JSON, we can be more direct and explicit about what this code is doing.

This is part of https://github.com/wellcomecollection/platform/issues/5449 – it's removing a link between the API repo and the display/pipeline models, which we eventually want to remove from this repo entirely.

### DIFF
--- a/search/src/test/scala/weco/api/search/json/CatalogueJsonUtilTest.scala
+++ b/search/src/test/scala/weco/api/search/json/CatalogueJsonUtilTest.scala
@@ -5,622 +5,123 @@ import io.circe.parser.parse
 import org.scalatest.EitherValues
 import org.scalatest.funspec.AnyFunSpec
 import org.scalatest.matchers.should.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks
 import weco.api.search.models.request.{WorkInclude, WorksIncludes}
-import weco.catalogue.display_model.test.util.DisplaySerialisationTestBase
-import weco.catalogue.internal_model.generators.ImageGenerators
-import weco.catalogue.internal_model.work.generators.{
-  ItemsGenerators,
-  ProductionEventGenerators,
-  SubjectGenerators,
-  WorkGenerators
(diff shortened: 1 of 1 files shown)


search/src/test/scala/weco/api/search/json/CatalogueJsonUtilTest.scala
```diff
@@ -5,622 +5,128 @@ import io.circe.parser.parse
...
-      val w = indexedWork().contributors(
-        List(
-          Contributor(agent = Agent(randomAlphanumeric(25)), roles = Nil)
-        )
-      )
-
-      it("includes contributors with WorkInclude.Contributors") {
-        val includes = WorksIncludes(WorkInclude.Contributors)
-        val json = w.asJson(includes)
+    it("includes/omits fields based on the work include") {
```
paul-butcher, line 113: (Similar to my comment on another #430)
Because this is refactoring and replicates what was already there, these aren't change requests, but a bit of work to consider for the future.

This test guarantees that the given fields will be removed from the top level if they are present.  However, it does not show that they will not be removed from a nested position.

When reading this change in isolation, it took a while to realise that that is why `identifiers` has its own test, distinct from all the other properties.

Also for complete coverage, I'd like to see 

* ensure that it removes more than one field when excluded
* ensure that it retains more than one field when included

As it is, the test does cover the remaining condition - 
* ensure that it still succeeds if there is nothing to do.

paul-butcher (approved):

search/src/test/scala/weco/api/search/json/CatalogueJsonUtilTest.scala
alexwlchan: I’ve added a couple of extra test cases and some comments to explain why `identifiers` are special.

PR #442, opened 2022-05-16
Look up works with items that are unidentified

alexwlchan:
This should fix the issue Natalie is seeing where she can't look up her requested items.

The ItemLookup asks the Catalogue API for all the works matching a given item ID, then inspects all the items on that work to see if they're the one it's looking for. If one of those items is unidentified, it's obviously not what we want, but it calls `item.identifiers.head` which throws a `NoSuchElementException: head of empty list` exception.

This patch updates the ItemLookup to be aware of items without identifiers, and not error when it finds it.


jamieparkinson (approved):
Excellent

PR #445, opened 2022-05-16
Surface errors that we don't understand

jamieparkinson:
There wasn't an obvious need for an ADT to represent different types of errors here - the only type I could bring to mind was this "this shouldn't happen but here we are" error

### DIFF
--- a/requests/src/main/scala/weco/api/requests/services/RequestsService.scala
+++ b/requests/src/main/scala/weco/api/requests/services/RequestsService.scala
@@ -57,14 +57,13 @@ class RequestsService(
   ): Future[List[(SierraHold, RequestedItemWithWork)]] =
     for {
       holdsMap <- sierraService.getHolds(patronNumber)
-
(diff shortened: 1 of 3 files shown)


requests/src/main/scala/weco/api/requests/services/RequestsService.scala
```diff
@@ -57,14 +57,13 @@ class RequestsService(
...
       holdsMap <- sierraService.getHolds(patronNumber)
-
       itemLookupResults <- itemLookup.bySourceIdentifier(holdsMap.keys.toSeq)
 
       itemsFound = itemLookupResults.zip(holdsMap.keys).flatMap {
         case (Right(item), _) => Some(item)
         case (Left(itemLookupError: ItemLookupError), srcId) =>
           error(s"Error looking up item $srcId.", itemLookupError.err)
-          None
+          throw itemLookupError.err
```
alexwlchan, line 66: Can we do away with the `Some` in the other arm?

jamieparkinson: yes

PR #446, opened 2022-05-17
Remove some now-unused dependencies on display model

alexwlchan:

More significantly, this gets internal model out of those applications entirely.


alexwlchan:
Ah, no, this won't work – we need IndexFixtures.

PR #453, opened 2022-05-19
Only fetch information for "stale" Sierra items

alexwlchan:

One of the remaining couplings between the pipeline/API repo is all the logic for deciding if a Sierra item is requestable (the rules for requesting, item access rules). This is duplicated across the repos and we have Buildkite checks that the two move in sync. This ties us both to the internal model and Scala specifically, which we want to eventually ditch in the items API. This dependency prevents us doing that.

I think we can drastically simplify this logic in the items API.

In particular: we know a lot of items are not going to change in a way that requires real-time updates on the site; we can accept the latency in the catalogue API. Only items that are requestable need real-time information. (We already have [logic in the front-end](https://github.com/wellcomecollection/wellcomecollection.org/blob/fbec553332d061a6cdec5580c591ca810833e629/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx#L33-L52) that will decide if an item is "stale". If an item isn't stale, we skip going to the items API.)

This patch starts by tweaking the logic in the Sierra item updater:

* If an item is stale, we get new data from Sierra and re-run the access condition rules
* If an item is up-to-date, we use the access condition from the catalogue API

In a subsequent PR, this will let us cut down the amount of access condition related logic.

### DIFF
--- a/items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
+++ b/items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
@@ -85,16 +85,19 @@ class SierraItemUpdater(sierraSource: SierraSource)(
     val staleItems = itemMap
       .filter { case (_, item) => isStale(item) }
 
-    debug(s"Asked to update items ${itemMap.keySet}, refreshing stale items ${staleItems.keySet}")
+    debug(
+      s"Asked to update items ${itemMap.keySet}, refreshing stale items ${staleItems.keySet}"
+    )
 
     for {
       accessConditions <- getUpdatedAccessConditions(staleItems.keys.toSeq)
 
-      updatedItems = itemMap.map { case (sierraId, item) =>
-        accessConditions.get(sierraId) match {
-          case Some(updatedAc) => updateAccessCondition(item, updatedAc)
-          case None            => item
-        }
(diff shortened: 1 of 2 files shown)


paul-butcher (approved):

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -88,26 +76,99 @@ class SierraItemUpdater(sierraSource: SierraSource)(
...
+    *      Another user can no longer request the item, but the catalogue API will
+    *      tell you it's available.
+    *
+    * In all other cases, we can use the access conditions in the catalogue API.
+    * There may be a small delay in updating an item's information, but this is
+    * relatively tolerable, and allows us to simplify the logic in this API for
+    * determining item status.
+    *
+    */
+  private def isStale(item: DisplayItem): Boolean = {
```
jamieparkinson, line 149: Do you want to add a link to https://github.com/wellcomecollection/wellcomecollection.org/blob/fbec553332d061a6cdec5580c591ca810833e629/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx#L25-L28 in the comments for bonus points?

jamieparkinson (approved):

PR #454, opened 2022-05-19
Don't ask Sierra about items we know are unrequestable

alexwlchan:

Right now this is a relatively small change to the requests API internals, but it will allow us to start removing some of the logic for access conditions we'll never encounter in practice. I'm going to do that in a separate PR, to keep the diffs manageable.

### DIFF
--- a/requests/src/test/scala/weco/api/requests/RequestingScenarioTest.scala
+++ b/requests/src/test/scala/weco/api/requests/RequestingScenarioTest.scala
@@ -18,10 +18,9 @@ import weco.catalogue.internal_model.generators.IdentifiersGenerators
 import weco.catalogue.internal_model.identifiers.CanonicalId
 import weco.http.client.{HttpGet, HttpPost, MemoryHttpClient}
 import weco.sierra.generators.SierraIdentifierGenerators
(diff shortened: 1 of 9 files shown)


requests/src/test/scala/weco/api/requests/RequestingScenarioTest.scala
```diff
@@ -1428,77 +1427,6 @@ class RequestingScenarioTest
     api.routes
   }
 
-  def createListHoldsRequest(patron: SierraPatronNumber): HttpRequest =
```
alexwlchan, line 1431: All this code has been moved unmodified to SierraServiceFixture, so it can be used in RequestingScenarioTest.

requests/src/test/scala/weco/api/requests/fixtures/SierraServiceFixture.scala
```diff
@@ -35,4 +37,75 @@ trait SierraServiceFixture extends HttpFixtures with Akka {
 
       testWith(sierraService)
     }
+
+  def createHoldRequest(
```
alexwlchan, line 51: All this code has been moved unmodified from RequestingScenarioTest, so it can be used in SierraRequestsServiceTest.

PR #458, opened 2022-05-20
Revert "Update date format in HoldNote.scala" #452

jamieparkinson:
We don't need this change any more, and it introduces an ambiguity into the dates


paul-butcher (approved):
ISO 8601 FTW!

PR #460, opened 2022-05-23
Trim the ItemAccess rules to what we actually use in this API

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5530; follows #456

This drastically simplifies the logic in the items and requests APIs (and all their existing tests still pass); because most of this code was covering cases we'll never hit in practice.

e.g. if an item has fixed field 88 = m / missing, it will be marked as "Unavailable" in the catalogue API. The front-end should never query the items API for the up-to-date status of this item (because it knows it won't change) or display a "Request item" button (because it knows it won't work). Having code to handle that case in these APIs is wasteful.

Now these rules no longer need to move in lockstep with the pipeline repo, I can update them to use just the display models and cut the link to internal model.


jamieparkinson (commented):
Not sure about this to be honest - isn't it increasing the amount of code we have to _maintain_ by making it diverge from the code in the pipeline repo?

alexwlchan:
> isn't it increasing the amount of code we have to maintain by making it diverge from the code in the pipeline repo?

Yes and no.

It lets us split off from internal model, the vendored code across repos (and eventually Scala entirely). Because it's a massively simplified version, I think we'll be touching it extremely rarely. Otherwise we end up with a coupling between pipeline and API (because some of this logic does belong in the pipeline imo).

alexwlchan:
Ah, hmm, I can think of another way to do this. Let me rethink this.

jamieparkinson (approved):

Issue #466, opened 2022-05-24
multi-word exact matches at the start of a title aren't always first in the list of search results


harrisonpim:
> Hi gang, search ranking question - I was trying to find the book "Information Law", by searching on those words:  https://wellcomecollection.org/works?query=information+law. As you can see the book doesn't show up until about half way down the second page of hits, even though the words are an exact match to the first two words of the title. Is that to be expected (because the book's works page isn't very popular or something?). If I add the author name to the search, then it shows up as the only hit (the book is https://wellcomecollection.org/works/zkg7xqm7).

harrisonpim:
the current query is prioritising matches in the `search.relations`, which has no positional awareness. we don't currently prioritise proximity to the start of a title (or any other field)

harrisonpim:
The 'correct' result for this query can be brought back by adding a [span_first](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-span-first-query.html) query:

```json
{
  "query": {
    "bool": {
      "should": [
        {
          "span_first": {
            "match": {
              "span_term": {
                "data.title.shingles": "information law"
              }
            },
            "end":1,
            "boost":1000
          }
        },
        {
          "multi_match": {
...
```

harrisonpim:
I'd like to investigate the effect of
- turning down the boost on the relations field vs
- adding the `span_first` to `title.shingles`
using the standard set of rank tools before making any changes to the prod query

pollecuttn:
Issue first reported on Slack at https://wellcome.slack.com/archives/C8X9YKM5X/p1653293886888919

PR #468, opened 2022-05-24
add a rank test for exact matches at the start of a work's title

harrisonpim:
Adding a test to rank for https://github.com/wellcomecollection/catalogue-api/issues/466. 

It's a `knownFailure` until we improve the query...


rank/data/tests/works.ts
```diff
@@ -11,6 +9,13 @@ const tests: Test[] = [
       'Ensure that the query brings back the works we expect at the top of the list',
     eval: equalTo1,
     cases: [
+      {
+        query: 'information law',
+        ratings: ['zkg7xqm7'],
+        knownFailure: true,
+        description:
+          'Multi-word exact matches at the start of a title should be prioritised',
+      },
```
alexwlchan, line 18: Can you toss in a comment linking to the original Slack discussion and/or GitHub issue, so we know this test comes from a real user?

jamieparkinson (approved):

PR #470, opened 2022-05-25
update the rank tests for french and italian stemming

### LINKED ISSUE #469: why is rank failing?

harrisonpim:
resolves #469


harrisonpim:
Doesn't feel great to be changing the tests without a good explanation of what _actually_ caused them to start failing, but we can see that italian and french works _are_ still being analysed appropriately. The tests still reflect that, and are no longer blocking further query development

jamieparkinson (approved):

Issue #459, opened 2022-05-20
Some items are on hundreds of works, breaking the requests API


jamieparkinson:
These items and presumably more are on many works:
`i18489102` - 289 works
`i18489096` - 249 works
`i18489126` - 245 works

The requests API only looks at the first 100 results when querying `/works` for a set of item identifiers: https://github.com/wellcomecollection/catalogue-api/blob/main/requests/src/main/scala/weco/api/requests/services/ItemLookup.scala#L124

This means that if a user has these items on hold, then they won't be able to see their requests, as the above linked method will fail. We should fix this: the obvious approach is to introduce paginated queries.

alexwlchan:
> the obvious approach is to introduce paginated queries

Not sure about “obvious”; it's definitely one approach, but potentially runs up dozens of redundant requests (I think there's an item on thousands of works). There are other approaches we could consider also:

* Make separate queries for each of the identifiers, or progressively cut the number of identifiers you query

   e.g. on query 1 I fetch `item.identifiers=A,B,C,D`. I find `A` and `B` and see totalResults is > 100, so on the next query I fetch `item.identifiers=C,D`. Repeat until an identifier can't be found.

* Start storing the work ID somewhere, so the requests API just looks up a list of N work IDs, as we've discussed previously.

jamieparkinson:
Don't think we really have the resource to do option 2 - option 1 is a nice optimisation, let's do that

PR #471, opened 2022-05-25
Users may have multiple items on hold spanning many works

jamieparkinson:
Resolves https://github.com/wellcomecollection/catalogue-api/issues/459

This fix looks at all the pages of a works search - not particularly efficient, but should do the job.


alexwlchan (approved):
Looks fine. The API does handle pagination for you, in the form of the `nextPage` parameter, but this approach is fine and I don't think it's worth changing now.

jamieparkinson:
Yeah, makes using the api client & logging a bit of a pain though

alexwlchan:
\*nod*

Yeah, I think you could do it quite nicely if you designed your whole API client around it, but we don't need to do that here.

PR #473, opened 2022-05-27
Catalogue-api is now a JS repo

jamieparkinson:
This repo contains the microservices which constitute the catalogue API. Currently these are all Scala projects, but we're going to add a JavaScript one next. 

We might convert others in future: this PR is a statement of intent. It creates a yarn workspace with prettier formatting on precommit.

In the process of doing this work I'm going to be adding things like the prettier config, and then perhaps linting and typescript config. We want to standardise these, but I think it would be doing it backwards to determine the standard first. After they've stabilised here, I'll put copies of them somewhere else so that we can use them for all Wellcome Collection JS projects.


alexwlchan (approved):
👀 

485 lines of dependency to do zilch. JavaScript!

harrisonpim:
✨

PR #475, opened 2022-05-27
Prioritise exact matches at the start of titles

harrisonpim:
This PR adds a [span_first](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-span-first-query.html) query to the `title.shingles` field. 

Adding this section to the query resolves https://github.com/wellcomecollection/catalogue-api/issues/466, allowing us to remove the `knownFailure` label from the "information law" rank test. The rest of the rank tests are unaffected by the change.

According to the `compareQuerySpeed` script, there's no significant performance cost to the addition either.

### DIFF
--- a/search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala
+++ b/search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala
@@ -42,7 +42,7 @@ case object WorksMultiMatcher {
         //
         //      Human genetic information : science, law, and ethics
         //      International journal of law and information technology
-        //      Information law : compliance for librarians and information professionals 
+        //      Information law : compliance for librarians and information professionals
         //
         // and somebody searches for "Information law", all other things being equal,
         // we want to prioritise the third result.
--- a/search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala
+++ b/search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala
@@ -12,6 +12,10 @@ import com.sksamuel.elastic4s.requests.searches.queries.matches.{
   MatchQuery,
   MultiMatchQuery
 }
+import com.sksamuel.elastic4s.requests.searches.span.{
+  SpanFirstQuery,
+  SpanTermQuery
+}
 
 case object WorksMultiMatcher {
   val titleFields = Seq(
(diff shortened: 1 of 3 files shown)


search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala
```diff
@@ -32,6 +36,15 @@ case object WorksMultiMatcher {
...
       .should(
+        SpanFirstQuery(
+          SpanTermQuery(
+            field = "data.title.shingles",
+            value = q
+          ),
+          boost = Some(1000),
+          queryName = Some("start of title"),
+          end = 1
+        ),
```
alexwlchan, line 47: I realise this is kinda handled by the rank tests, but let's put a comment explaining what this is doing.

```
// This prioritises exact matches at the start of titles.
//
// e.g. if we had three works
//
//      Human genetic information : science, law, and ethics
//      International journal of law and information technology
//      Information law : compliance for librarians and information professionals 
//
// and somebody searches for "Information law", all other things being equal,
// we want to prioritise the third result.
//
// See https://github.com/wellcomecollection/catalogue-api/issues/466
```

Among other things, capturing some examples is helpful because we can only see the "before" right now; once we fix this the bad behaviour in the ticket will disappear.

harrisonpim: Happy to add this temporarily, but in general I'd like to avoid documenting every decision everywhere. I know that search documentation is lacking right now, but I think its scattered nature is part of the problem... It'll be easier to read (and write) if it's managed in one canonical location. All this to say, this comment ☝️ might be bounced over to gitbook when I tackle that problem properly next week.

alexwlchan: If there's a canonical record of these decisions somewhere that can be linked to from the query definition, that works for me too. 😄

alexwlchan (approved):

search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala
alexwlchan: Happy for this to go in as-is without the comment if you're doing search documentation next week, and consider that an example of what I'd want to see in there.

PR #477, opened 2022-05-30
Use Docker images from ECR Public, not our vendored repos

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5545, which includes a discussion of the benefits.


jamieparkinson (approved):
Interested to see if the higher rate limits that apply to pulls from within AWS (ie our CI) Just Work here

alexwlchan:
🤞 

I'm hoping simplifying the number of different images we might help with that; any given instance only has to pull, say, a single Node 12 image, rather than four slightly different Node 12 images.

PR #481, opened 2022-05-30
Build/test/push the concepts API in CI

jamieparkinson:
lemon squeezy

Only interesting bit is the change to `run_job.py` to avoid re-building scala projects when this mysterious javascript project changes.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -47,8 +47,21 @@ steps:
   - label: "Test concepts API"
     plugins:
       - docker-compose#v3.9.0:
-          run: concepts_api
+          run: concepts
           command: ["yarn", "test"]
+
+  - label: "Publish concepts API (ECR)"
+    #    if: build.branch == "main"
+    if: build.branch == "concepts-ci"
+    plugins:
+      - wellcomecollection/aws-assume-role#v0.2.2:
+          role: "arn:aws:iam::756629837203:role/catalogue-ci"
+      - ecr#v2.1.1:
+          login: true
+      - docker-compose#v3.9.0:
+          push:
+            - concepts:756629837203.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/concepts:ref.${BUILDKITE_COMMIT}
+            - concepts:756629837203.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/concepts:latest
 
   - command: .buildkite/scripts/run_job.py snapshot_generator
     label: "Snapshot generator"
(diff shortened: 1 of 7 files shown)


alexwlchan:
> Only interesting bit is the change to run_job.py to avoid re-building scala projects when this mysterious javascript project changes.

Q: Should we do away with that logic entirely? The entire build runs in <5m, is it worth trading a simpler build system for more EC2 minutes?

jamieparkinson:
Not at the moment I don't think - let's just wait until we get rid of scala in this repo 😉

.buildkite/pipeline.yml
```diff
@@ -44,6 +44,27 @@ steps:
...
       queue: "scala"
 
+  - label: "Test concepts API"
+    key: "concepts-test"
+    plugins:
+      - docker-compose#v3.9.0:
+          run: concepts
+          command: ["yarn", "test"]
+
+  - label: "Publish concepts API (ECR)"
```
alexwlchan, line 54: I realise this is probably copied from the dotorg repo; could we use consistent labelling with the other tasks in this repo please?

alexwlchan:
Okay, let's leave it for now, but I think we should consider it across the board for our Scala apps. (Possibly coupled with a breaking up of the pipeline repo.)

terraform/shared/ecr.tf
```diff
@@ -33,6 +33,14 @@ resource "aws_ecr_repository" "api" {
   }
 }
 
+resource "aws_ecr_repository" "concepts" {
+  name = "uk.ac.wellcome/concepts"
```
alexwlchan, line 37: weco/concepts?

Or does that wreak havoc with weco-deploy?

alexwlchan: Also, is this an opportunity to start defining some better defaults around ECR repository policies for expiring old images?

.buildkite/pipeline.yml
jamieparkinson: It wasn't - that's not a straightforward thing to do as (1) the other tasks handle the test/push logic inside `run_job.py` and (2) they're very inconsistently named as it is. I've renamed some things a little but there is no way of doing this consistently with the other APIs as the steps aren't exactly analogous.

terraform/shared/ecr.tf
jamieparkinson: Yeah, this naming is to satisfy weco-deploy unfortunately.

I was thinking about repo policies too - yes, we should do that, but not as part of this PR. Standardising the policies won't be harder if we have an extra repo, and I don't want this effort to get a service up and running quickly to accrue wider improvements as we do it. Rather, we should be tracking the things that it makes us notice and then fixing them as we see fit: https://github.com/wellcomecollection/platform/issues/5546

alexwlchan: Sounds good on both counts. 👍

alexwlchan (approved):

PR #482, opened 2022-05-30
Add concepts service to the catalogue API

jamieparkinson:
This was ridiculously straightforward!

### DIFF
--- a/terraform/modules/stack/services.tf
+++ b/terraform/modules/stack/services.tf
@@ -102,6 +102,11 @@ module "concepts_api" {
 
   secrets = {}
 
+  // TODO increase these when this is a production service
+  // These are the minima allowed by Fargate
+  app_cpu    = 256
+  app_memory = 512
+
   # Below this line is boilerplate that should be the same across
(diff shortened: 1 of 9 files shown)


terraform/modules/stack/services.tf
```diff
@@ -81,3 +82,26 @@ module "items_api" {
...
+  desired_task_count          = var.desired_task_counts.concepts
+  load_balancer_listener_port = local.concepts_lb_port
+  security_group_ids          = local.ecs_security_groups
+
+  environment = {
+    PORT = local.container_ports.concepts
+  }
+
+  secrets = {}
+
```
alexwlchan, line 101: I assume this is true; if not, we should fix it!

```suggestion
  # Below this line is boilerplate that should be the same across
  # all Fargate services.
```

alexwlchan (approved):
Looks good for getting it up and running; I wonder if we can turn down the CPU/memory on it for now? To avoid adding 4x API tasks (prod+staging) for a prototype.

PR #487, opened 2022-05-31
Put an elasticsearch client in the concepts app

jamieparkinson:
Yet to be used but gets some of the boring bits of https://github.com/wellcomecollection/catalogue-api/issues/486 out of the way.

The dependency injection pattern used here (`createApp(clients)`) worked well for the identity API and we'll easily be able to inject a mock client for testing as per https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/client-testing.html

### DIFF
--- a/concepts/src/services/aws.ts
+++ b/concepts/src/services/aws.ts
@@ -12,7 +12,7 @@ export const getSecret = async (id: string): Promise<string | undefined> => {
     );
     return result.SecretString;
   } catch (e) {
-    console.error(`Error fetching secret '${id}`, e);
+    console.error(`Error fetching secret '${id}'`, e);
     return undefined;
   }
 };


concepts/src/app.ts
```diff
@@ -1,6 +1,31 @@
 import express from "express";
+import type { Client as ElasticClient } from "@elastic/elasticsearch";
 
-const app = express();
+type Clients = {
+  elastic: ElasticClient;
```
alexwlchan, line 5: Just to check we're on the same page: this structure would allow us to add, say, a `SierraClient` when we do the items and requests APIs?

concepts/src/app.ts
```diff
@@ -1,6 +1,31 @@
...
+      results: Object.values(concepts),
+    });
+  });
+
+  app.get("/concepts/:id", (req, res) => {
+    const concept = concepts[req.params.id];
+    if (concept) {
+      res.status(200).json(concept);
+    } else {
+      res.sendStatus(404);
```
alexwlchan, line 23: We certainly don't need it for 1.0, but seeing this makes me wonder if we'll ever need to support 302 Redirect/410 Gone in this API.

concepts/src/services/aws.ts
```diff
@@ -0,0 +1,18 @@
...
+const client = new SecretsManagerClient({});
+
+export const getSecret = async (id: string): Promise<string | undefined> => {
+  try {
+    const result = await client.send(
+      new GetSecretValueCommand({ SecretId: id })
+    );
+    return result.SecretString;
+  } catch (e) {
+    console.error(`Error fetching secret '${id}`, e);
```
alexwlchan, line 15:
```suggestion
    console.error(`Error fetching secret '${id}'`, e);
```

alexwlchan (approved):

PR #490, opened 2022-06-06
Use the new query fields in the search API

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5550; accompanies https://github.com/wellcomecollection/catalogue-pipeline/pull/2107

This is a much smaller change than the diff implies; most of it is copying across autogenerated test documents. All the interesting stuff is in the final commit.

This repo is now totally divorced from `WorkState.Indexed`; it no longer uses it. We can't get rid of it quite yet because it's used for some copyTo in the Elasticsearch mapping, but we're getting closer to removing it.

### DIFF
--- a/common/search/src/test/resources/WorksIndexConfig.json
+++ b/common/search/src/test/resources/WorksIndexConfig.json
@@ -952,6 +952,14 @@
       "query": {
         "type": "object",
         "properties": {
+          "id" : {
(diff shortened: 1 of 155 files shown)


alexwlchan:
I've kicked off a reindex to populate the index with the fields this is looking at, but I can't merge it until that's done.

PR #491, opened 2022-06-06
Prototype Concepts API

jamieparkinson:
Pure Elasticsearch hackery. Includes some actually-useful/scalable stuff for error handling and pagination, though.


concepts/server.ts
```diff
@@ -1,11 +1,13 @@
 import createApp from "./src/app";
 import { getElasticClient } from "./src/services/elasticsearch";
 
+const pipelineDate = "2022-05-30";
+
 getElasticClient({
-  serviceName: "concepts_api",
-  pipelineDate: "2022-05-30",
+  serviceName: "catalogue_api", // TODO an elasticsearch user for the concepts API
```
alexwlchan, line 7: I assume this is blocked on provisioning ES users in TF or similar?

concepts/src/controllers/concept.ts
```diff
@@ -0,0 +1,53 @@
...
+      body: {
+        size: 1,
+        _source: ["display.subjects"],
+        query: {
+          term: {
+            "query.subjects.id": id,
+          },
+        },
+      },
+    });
```
alexwlchan, line 30: Doing this with Elasticsearch is fine now you've done it, but FWIW, my initial plan was to proxy responses from the public API (hence why I added the subject ID filtering).

concepts/src/controllers/concept.ts
```diff
@@ -0,0 +1,53 @@
...
+          });
+          return;
+        }
+      }
+    }
+
+    throw new HttpError({
+      status: 404,
+      label: "Not Found",
+      description: `Concept not found: ${id}`,
```
alexwlchan, line 49: nit: for consistency with the existing works/images APIs.

```suggestion
      description: `Concept not found for identifier ${id}`,
```

concepts/src/controllers/error.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const errorHandler: ErrorRequestHandler = (err, req, res, next) => {
+  const httpError =
+    err instanceof HttpError
+      ? err
+      : new HttpError({
+          status: 500,
+          label: "Server error",
+          description: "Expected server error: " + err.toString(),
+        });
```
alexwlchan, line 52: Minor: past convention has been to omit the `description` on 500 Internal Server Error, because from a user's POV we can't say anything useful there. I don't mind changing it, but want to check it's deliberate.

concepts/src/controllers/pagination.ts
```diff
@@ -0,0 +1,102 @@
...
+  totalResults: number;
+  prevPage?: string;
+  nextPage?: string;
+};
+
+export const paginationElasticBody = ({
+  page,
+  pageSize,
+}: PaginationQueryParameters): { from: number; size: number } => {
+  const size = Math.min(pageSize ?? defaultPageSize, limits.maxSize);
```
alexwlchan, line 28: I think this is fine for a hacky POV, but can we get a TODO in here that eventually we'll send users a 400 pointing them to snapshots instead?

concepts/src/controllers/pagination.ts
```diff
@@ -0,0 +1,102 @@
...
+const parsePaginationQueryParameters = (
+  params: URLSearchParams
+): PaginationQueryParameters => {
+  const page = parseNumberParam("page", params);
+  const pageSize = parseNumberParam("pageSize", params);
+
+  if (page !== undefined && page < 1) {
+    throw new HttpError({
+      status: 400,
+      label: "Bad request",
```
alexwlchan, line 44: nit: these are usually capitalised

```suggestion
      label: "Bad Request",
```

concepts/server.ts
jamieparkinson: Indeed

concepts/src/controllers/concept.ts
jamieparkinson: Yeah, I realised that after I'd done it and am not sure why I thought ES was necessary - on the other hand, this does make iteration a bit easier.

concepts/src/controllers/pagination.ts
jamieparkinson: [We do actually send a 400](https://github.com/wellcomecollection/catalogue-api/blob/0f4f4908e8b59430c8c76419ec84f57c279e8c94/concepts/src/controllers/pagination.ts#L52-L56) with the same message as the prod APIs, but agreed that it should point to the snapshots...

concepts/src/controllers/error.ts
jamieparkinson: This should also say "unexpected" 😬

PR #497, opened 2022-06-15
Use the new index structure for work/image filters

alexwlchan:
This will need a docs change when it's live.


alexwlchan:
Replaced by #498

PR #499, opened 2022-06-17
add a script to run a search with a candidate query/index from the rank cli

harrisonpim:
I'm using the rank web app less and less these days, as the CLI wrappers around the rank services have matured. I'd like to gradually phase out the web UI completely.

This PR adds a script to display rendered search results from candidate queries from the CLI, replicating the functionality on  the `/search` page of the web app.

I've also tidied up a couple of other little pieces:
- the `updateIndex` script now prompts the user to run an `updateByQuery` on the affected index
- `checkReindex` can now check reindexes and updateByQuery tasks, and has been renamed accordingly
- including `explain` results in the search service is now optional


alexwlchan:
> I'd like to gradually phase out the web UI completely.

Is this the bit that's hosted on Vercel?

(Motivation: Vercel is a bit of an oddity in our deployment story, and if you don't want to use it I'd be tempted to push ahead on getting rid of it entirely, at least until we have another use of it.)

rank/scripts/search.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const results = await search({ index, queryEnv, searchTerms, explain: false })
+
+  results.hits.hits.map((hit, i) => {
+    const title = hit._source.display.title
+    const url = `https://wellcomecollection.org/works/${hit._id}`
+    const reference = hit._source.display.referenceNumber
+    const dates = hit._source.display.production
+      .flatMap((p) => p.dates.map((d) => d.label))
+      .join(', ')
```
alexwlchan, line 36: The fact that you can get this from the `display` object and trust it not to change is quite nice. 🥳

rank/scripts/search.ts
```diff
@@ -0,0 +1,49 @@
...
+    const displayString = [
+      `${i + 1}.\t${chalk.bold(chalk.cyan(title))}`,
+      reference ? `${reference}` : null,
+      dates ? `${dates}` : null,
```
alexwlchan, line 40: TypeScript novice question: is there a reason you need to template these as strings? I'd expect this to be equivalent to:

```suggestion
      reference ? reference : null,
      dates ? dates : null,
```

What am I missing?

I also wonder if you could just use `references` and `dates` directly here, given you're filtering further down.

rank/scripts/updateIndex.ts
```diff
@@ -84,6 +84,26 @@ async function go() {
...
+    const { body: updateByQueryRes } = await rankClient.updateByQuery({
+      wait_for_completion: false,
+      index: indexName,
+      conflicts: 'proceed',
+    })
+
+    const taskID = updateByQueryRes.task
+    success(`Update started successfully with task ID ${taskID}`)
+    info('You can monitor the task by running:')
+    code('yarn checkTask')
```
alexwlchan, line 105: Not for this patch perhaps, but this seems to lend itself to passing a task ID here, e.g.

```
yarn checkTask 289c9e59-318b-42cc-9fd4-763fb7e506a8
```

so you can inspect your task specifically, rather than finding it in the list of all tasks in the cluster.

alexwlchan (approved):
🍎

rank/scripts/search.ts
harrisonpim: Oops, you're right! I think that's a hangover from an intermediate version of this section which never got scrubbed

rank/scripts/updateIndex.ts
harrisonpim: I think you're right here too - the automatic task discovery thing is nice, but users should really be able to specify the task they're checking. Maybe the best-of-both-worlds is to allow a task ID to be specified, but if it's not, give the user a choice from all available tasks.

I'll tidy this up in a separate PR.

harrisonpim:
> > I'd like to gradually phase out the web UI completely.
> 
> Is this the bit that's hosted on Vercel?
> 
> (Motivation: Vercel is a bit of an oddity in our deployment story, and if you don't want to use it I'd be tempted to push ahead on getting rid of it entirely, at least until we have another use of it.)

The knowledge graph prototype is also currently running on vercel, but I can imagine that also being phased out for the prototype APIs and pages being developed within the core pipeline soon.

I still think vercel's a useful way of getting things up and running quickly, and we might yet have use for it with other prototype-y, demo-y bits of work. Definitely worth keeping this discussion going though.

PR #500, opened 2022-06-18
Use the 2022-06-18 index

alexwlchan:
Accompanies https://github.com/wellcomecollection/catalogue-pipeline/pull/2131; closes https://github.com/wellcomecollection/platform/issues/5561

In particular, images in the index now have source data from the primary work, and not the canonical/redirected work separately. This slightly simplifies the queries we need to make in the API.

### DIFF
--- a/search/src/main/scala/weco/api/search/services/WorksService.scala
+++ b/search/src/main/scala/weco/api/search/services/WorksService.scala
@@ -60,6 +60,7 @@ class WorksService(val elasticsearchService: ElasticsearchService)(
       case Right(resp) =>
         Right(
           resp.aggregations
+//            .getAgg("work_type")
             .decodeAgg[String]("work_type")
             .get
             .buckets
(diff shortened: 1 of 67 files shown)


search/src/main/scala/weco/api/search/services/WorksService.scala
```diff
@@ -60,6 +60,7 @@ class WorksService(val elasticsearchService: ElasticsearchService)(
       case Right(resp) =>
         Right(
           resp.aggregations
+//            .getAgg("work_type")
```
paul-butcher, line 63: Is this going to be uncommented soon?

alexwlchan: Oops, should be removed.

alexwlchan, line 63:
```suggestion
```

paul-butcher (approved):

PR #505, opened 2022-06-21
Filter subject/genre labels from the query fields

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5550; I think these fields went in on the last reindex, and I just never switched to using them in the API.


paul-butcher:
It looks like genre and subject filters are failing on buildkite for this now.  Is this a mismatch between index and API that just has to wait for the next reindex?

alexwlchan:
Either this is wrong, or I need to copy across new test documents I think.

alexwlchan:
Closing as superseded.

Issue #488, opened 2022-05-31
rework the `search.relations` field and the way it's queried


harrisonpim:
The relations field was introduced to provide archive search with both IDs and natural language search terms. The current form of the relations field/query isn't necessarily the best way of achieving that goal, and it's beginning to spill out and interfere with other relevance problems. I suspect it's trying to do too many things at once. It should be restructured, making each aspect more explicit and easier to document.

harrisonpim:
relations should only match if the path is matched?

harrisonpim:
a bunch of useful fields currently use a non-existent `english` analyzer, instead of the correct `english_analyzer`

harrisonpim:
query smarter with a structure like [this](https://stackoverflow.com/questions/72000841/elasticsearch-match-all-search-tokens-in-one-or-any-fields-in-any-order) across the fields more deliberately, rather than overloading one new field with `copyTo`s

PR #509, opened 2022-06-27
Relations query

harrisonpim:
Replaces the blunt match query on `search.relations` with some smarter logic across the relevant relations path fields, and commonly queried text fields.

Related to https://github.com/wellcomecollection/catalogue-pipeline/pull/2135

Closes https://github.com/wellcomecollection/catalogue-api/issues/488

### DIFF
--- a/rank/scripts/createIndex.ts
+++ b/rank/scripts/createIndex.ts
@@ -31,7 +31,12 @@ async function go() {
     (mod) => mod.default
   )
 
-  const remoteIndex = `${getNamespaceFromIndexName(localIndex)}-candidate`
+  const remoteIndex = await prompts({
+    type: 'text',
+    name: 'value',
+    message: 'What should the new index be called?',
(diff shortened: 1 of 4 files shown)


rank/scripts/createIndex.ts
```diff
@@ -82,7 +87,7 @@ async function go() {
 
     success('Reindex started successfully')
     info('You can monitor the reindex by running:')
-    code(`yarn checkTask --task ${reindexResp.task.task_id}`)
+    code(`yarn checkTask --task ${reindexResp.task}`)
```
paul-butcher, line 90: I trust that this works, but I can't see how this has changed so that `.task_id` is no longer needed

harrisonpim: Turns out `task_id` was an incorrect addition the first time round 😅 It didn't work previously, and now it does!

paul-butcher (approved):

PR #526, opened 2022-08-12
Add threshold image similarity scores

### LINKED ISSUE #516: add a threshold score for visually similar images
Following https://github.com/wellcomecollection/platform/issues/5581, we think we should add a blunt threshold similarity score to remove spurious matches from the list of results. At a glance, 300 or 400 seems like a reasonable level

This is just a simple patch over a deeper problem, for which we'll be testing solutions next quarter

harrisonpim:
This PR sets a threshold similarity score for images so that we can avoid displaying questionable matches. Closes #516 

This change messes with some of our existing tests - the dummy index populated while setting up the tests can't produce the same scores as the fully populated prod index, and the thresholds therefore filter out any results which might have been matched. In some tests I've been able to manually set the `minScore` to 0, but in others which call the API itself, that's not possible.  I feel like these things might be more appropriately tested in `rank`, where tests can be run against a full index and scoring can be properly examined.

### DIFF
--- a/search/src/main/scala/weco/api/search/services/ImagesService.scala
+++ b/search/src/main/scala/weco/api/search/services/ImagesService.scala
@@ -42,7 +42,8 @@ class ImagesService(
   def retrieveSimilarImages(
     index: Index,
     imageId: String,
-    similarityMetric: SimilarityMetric = SimilarityMetric.Blended
+    similarityMetric: SimilarityMetric = SimilarityMetric.Blended,
+    minScore: Option[Double] = None
   ): Future[List[IndexedImage]] = {
     val builder = similarityMetric match {
       case SimilarityMetric.Blended =>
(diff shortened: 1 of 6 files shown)


search/src/main/scala/weco/api/search/services/ImagesService.scala
```diff
@@ -53,7 +54,16 @@ class ImagesService(
         requestBuilder.requestWithSimilarColors
     }
 
-    val searchRequest = builder(index, imageId, nVisuallySimilarImages)
+    val defaultMinScore: Double = similarityMetric match {
+      case SimilarityMetric.Blended  => 300
+      case SimilarityMetric.Features => 300
+      case SimilarityMetric.Colors   => 20
+    }
```
alexwlchan, line 61: Can you explain how you came up with these default scores? Why is `Colors` so much lower?

harrisonpim: Sure. There's [a notebook in the data science repo](https://github.com/wellcomecollection/data-science/blob/main/notebooks/similarity_problems/notebooks/01-similarity-scores.ipynb) which produced the results in https://github.com/wellcomecollection/platform/issues/5581. Based on that analysis, we decided that ~300 seemed like an appropriate threshold for the blended similarity metric.

I re-ran that analysis with the `state.inferredData.lshEncodedFeatures` and `state.inferredData.palette` fields individually, and produced these corresponding graphs:

**lshEncodedFeatures:**
2c12fc1b-c57d-490f-83fd-7590149831de

**palette:**
9c4a9ae3-49d1-4834-a769-f56f32e1bc1c

If my mental maths is right, the scores for colours are generally lower because the `state.inferredData.palette` field contains fewer, more commonly occurring terms.

alexwlchan: Ah, super. Maybe just include a comment pointing to that ticket, so we can find this again in future?

alexwlchan (approved):

PR #540, opened 2022-08-30
[WIP] Deployments sans weco-deploy

alexwlchan:
This is an initial stab at https://github.com/wellcomecollection/platform/issues/5597, in particular the bit of weco-deploy that deploys a set of images based on an ECR tag.

The rough process:

1. For the given services, look up what container images you expect to be using. This gives us a sort of "service spec":

    ```json
    {
        "task_definition": "arn:aws:ecs:eu-west-1:756629837203:task-definition/matcher:4",
        "containers": {
            "app": {
                "uri": "ecr.public.123/matcher:env.prod",
                "digest": "sha256:1234567890"
            },
            "logging": {
                "uri": "ecr.public.123/logging:env.prod",
                "digest": "sha256:0987654321"
            }
        }
    }
    ```

2. Look to see if all the running tasks match the service spec. If so, we're done!

3. If not, trigger a deployment of all the stale services, then keep looking for matching service specs until we're done.

Example usage, for deploying the staging API:

```
python3 .buildkite/scripts/ensure_ecs_tasks_are_using_latest_images.py \
  --cluster=catalogue-api-2021-04-26 \
  --services='stage-items-api, stage-search-api, stage-concepts-api' \
  --role-arn=arn:aws:iam::756629837203:role/catalogue-developer
```

Things that weco-deploy does that this doesn't:

* Anything involving tagging; this new script is driven entirely by image digests. This is simpler, and I think it gets us the "don't deploy stuff that hasn't changed" behaviour for free, so to speak. Less abstraction!
* Allow you to configure the interval for checking and the eventual failure of the task; I don't think we care about the interval and I'm planning to lean on Buildkite timeouts for timing out the whole process.
* Coloured output.

### DIFF
--- a/.buildkite/scripts/ensure_ecs_tasks_are_using_latest_images.py
+++ b/.buildkite/scripts/ensure_ecs_tasks_are_using_latest_images.py
@@ -203,20 +203,18 @@ def check_if_tasks_match_expected_containers(sess, *, cluster, expected_containe
             continue
 
         actual_task_containers = {
-            c['name']: {
-                'uri': c['image'],
-                'digest': c['imageDigest']
-            }
-            for c in t['containers']
+            c["name"]: {"uri": c["image"], "digest": c["imageDigest"]}
+            for c in t["containers"]
         }
 
         if actual_task_containers != expected_task_containers:
             result[service].append(
-                f'Incorrect containers in task {task_id}.\n'
-                f'Expected:\n' +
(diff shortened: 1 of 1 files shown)


alexwlchan:
This is a bit too big and complicated to be a copy-pasta script, so I'm going to look at incorporating this into weco-deploy.

PR #547, opened 2022-09-26
Update to the latest test documents and index mapping

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5550


alexwlchan:
I think the churn is caused by the addition of `primary: true` on contributors and the image aspect ratio churning the seed.

PR #553, opened 2022-09-28
add a script to populate a local .env file

harrisonpim:
We've deleted the web UI for rank, and the associated vercel project along with it. The vercel project/cli allowed us to populate a local `.env` file for rank with a single command, which made the setup process super smooth.

This PR replaces that functionality with a bash script.


alexwlchan:
I’m not a super fan of having local files with plaintext credentials in them, and I'd like us to move away from it; how would you feel about updating the rank CLI to fetch secrets directly from AWS every time?

PR #555, opened 2022-09-30
Exclusively use fields on the query object when searching Works

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5550; follows https://github.com/wellcomecollection/catalogue-pipeline/pull/2214 and https://github.com/wellcomecollection/catalogue-pipeline/pull/2216.

After this patch, everything work-related is being fetched from the `display`, `query`, or `type` fields.

There's a non-trivial change here: I've stopped using the copyTo fields:

* `search.identifiers` – this was completely redundant after the query restructuring. The clause in the WorksMultiMatcher query that used it was also searching all the fields that get copied into this field.
* `search.titlesAndContributors` – once I'd ditched `search.identifiers`, this felt like a weird standout, so I replaced it with searching the individual fields. IMO this will allow us to simplify the index mapping.

This may have implications for query performance/behaviour, so I’d like @harrisonpim to look at this when he’s back on Monday. I think it's fine, but I don't know why we do it this way and there might be a good reason to keep the old behaviour. (If we do, should we lean into it more and _only_ search `search.identifiers`/`query.allIdentifiers`?)

### DIFF
--- a/search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala
+++ b/search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala
@@ -36,7 +36,7 @@ case object WorksMultiMatcher {
     boost: Int,
     fields: Seq[String]
   ): Seq[FieldWithOptionalBoost] =
-    fields.map(FieldWithOptionalBoost(_, Some(boost.toDouble)))
+    fields.map(FieldWithBoost(_, boost))
 
   def apply(q: String): BoolQuery =
     boolQuery()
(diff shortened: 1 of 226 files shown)


search/src/main/scala/weco/api/search/elasticsearch/WorksMultiMatcher.scala
```diff
@@ -83,25 +91,31 @@ case object WorksMultiMatcher {
...
-              ).map(
-                language =>
-                  FieldWithOptionalBoost(
-                    s"search.titlesAndContributors.$language",
-                    None
+              fields = titleAndContributorFields.flatMap(
+                field =>
+                  nonEnglishLanguages.map(
+                    language =>
+                      FieldWithOptionalBoost(
```
paul-butcher, line 115: Given the frequency with which `FieldWithOptionalBoost(???, None)` is used, I wonder if it would be cleaner to have a `Field` object that does not take a boost argument.

alexwlchan: Good suggestion, I've added `FieldWithoutBoost`.

alexwlchan:
Because Harrison is off sick today, I've backed out the changes to `query.allIdentifiers` and `query.titlesAndContributors`. I don't want to block this work until he's feeling better, and that change is something we can drop in later.

jamieparkinson (approved):

PR #556, opened 2022-10-03
use the 2022-10-03 pipeline and new fields

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5550


alexwlchan:
I forgot to index image locations in the `query` object 🙈

search/src/main/scala/weco/api/search/elasticsearch/ImagesQuery.scala
```diff
@@ -40,56 +43,42 @@ case object ImageSimilarity {
...
-      "id.canonicalId",
-      "id.sourceIdentifier.value",
-      "data.otherIdentifiers.value"
-    )
-  ) ++ Seq(
-    "state.canonicalId",
-    "state.sourceIdentifier.value"
-  ).map(FieldWithoutBoost(_))
+  private val idFields: Seq[FieldWithOptionalBoost] =
+    Seq(
```
paul-butcher, line 54: There are only four fields here, where there were five.  I assume that's because id.canonicalId and state.canonicalId are duplicates?

search/src/test/scala/weco/api/search/models/QueryConfigTest.scala
```diff
@@ -63,14 +63,5 @@ class QueryConfigTest
         result.paletteBinSizes shouldBe QueryConfig.defaultPaletteBinSizes
       }
     }
-
-    it("returns the default config if the data is not found") {
```
paul-butcher, line 67: Is this no longer desirable?  Should this be replaced with a test for the desired behaviour in this scenario, or is it now impossible to reach?

search/src/main/scala/weco/api/search/elasticsearch/ImagesQuery.scala
alexwlchan: No, it’s `id.sourceIdentifier.value` and `data.otherIdentifiers.value`, which are two work fields that collapsed into `query.identifiers` – although we distinguish between source/other identifiers in the pipeline, whenever we search we want both, so I collapsed them into a single list.

```
source.id.canonicalId              ~> query.source.id
source.id.sourceIdentifier.value   ~> query.source.identifiers.value
source.data.otherIdentifiers.value ~> query.source.identifiers.value
state.canonicalId                  ~> query.id
state.sourceIdentifier.value       ~> query.sourceIdentifier.value
```

search/src/test/scala/weco/api/search/models/QueryConfigTest.scala
alexwlchan: Impossible to reach – this is from when inferredData was an optional field on images and so might be `None`, but in practice we know it's always going to be populated. See https://github.com/wellcomecollection/catalogue-pipeline/pull/2219

paul-butcher (approved):
just two questions, otherwise LGTM

PR #562, opened 2022-10-28
update indexedTime source to come from debug

melanierogan:
This should cover step ~one~ three from https://github.com/wellcomecollection/platform/issues/5621

## Pull request checklist

Possibly to indicate we're using debug for the indexedValue


jamieparkinson (approved):
Is this not step (3) of the linked ticket?

melanierogan:
> Is this not step (3) of the linked ticket?

it _is_ - thanks Jamie

PR #565, opened 2022-11-03
Ditch the clever "should we build this" logic

alexwlchan:
We have a bunch of complex build logic that decides to skip building things if we don't think they've changed – this comes from the days of Travis CI, when CI time was much more constrained.

It's not something we'd do if we started from fresh today – and indeed, the concepts-pipeline is a comparable project and has none of this.

This PR deletes all that "clever" logic, in favour of a "build everything all the time" approach. I think this will be more reliable and easier to understand in the long run.


jamieparkinson (approved):

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/5626

PR #567, opened 2022-11-04
We don't need to write dependencies in .sbt_metadata

alexwlchan:
A further simplification for https://github.com/wellcomecollection/platform/issues/5626

Plus fixing the broken build on main – we used to tag commits as `latest` so we had something to compare to on the next Buildkite build, but now we build everything that's no longer necessary.


jamieparkinson (approved):
You can/should also remove https://github.com/wellcomecollection/catalogue-api/blob/main/project/plugins.sbt#L14

PR #569, opened 2022-11-04
Deploy images without using weco-deploy

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5626

* Reduce the duplication in the Buildkite config using `matrix` (and in particular, some of the repetition introduced in previous patches). This is mostly a convenience for anybody adding new tasks.

* Manage images into ECR and ECS using a couple of bash scripts. Most of the heavy lifting is being done by the AWS CLI, and the line additions are just comments explaining how the scripts work and what they're doing.


.buildkite/pipeline.deploy-stage.yml
```diff
@@ -1,50 +1,36 @@
...
+    command: |
+      ENV_TAG="env.staging" .buildkite/scripts/update_ecr_image_tag.sh \
+        uk.ac.wellcome/concepts
+        uk.ac.wellcome/items
+        uk.ac.wellcome/search
+
+      CLUSTER="catalogue-api-2021-04-26" .buildkite/scripts/redeploy_ecs_services.sh \
+        stage-concepts-api \
+        stage-items-api \
+        stage-search-api
```
alexwlchan, line 13: The repetition is a bit annoying here, and we could maybe collapse these into a single pipeline, but this should work and I don't want to make too many more changes right now.

.buildkite/scripts/redeploy_ecs_services.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+<<EOF
+Redeploy an ECS service.
```
jamieparkinson, line 3: What's the difference between deploying and redeploying?

.buildkite/scripts/redeploy_ecs_services.sh
```diff
@@ -0,0 +1,53 @@
...
+  aws ecs update-service \
+    --cluster "$CLUSTER" \
+    --service "$serviceName" \
+    --force-new-deployment
+done
+
+for serviceName in "$@"
+do
+  echo "Waiting for $serviceName to be stable"
+  aws ecs wait services-stable \
```
jamieparkinson, line 50: Does this log stuff while it's waiting?

.buildkite/scripts/run_scala_task.sh
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env bash
+<<EOF
+Run a Scala task in Buildkite.  This includes:
+
+  - running tests
+  - building a Docker image (if applicable)
+  - publishing the Docker image to ECR (if on main)
```
jamieparkinson, line 7: This seems like it does a lot more than just running a scala task! Can it be split up or renamed to make it clear what it does? Can the conditional (particularly re publishing on main) be put in the Buildkite pipeline instead? I know that the latter means the image needs to be rebuilt (as it's a new task) but I wonder if that tradeoff is OK.

.buildkite/scripts/run_scala_task.sh
```diff
@@ -0,0 +1,126 @@
...
+    PROJECT_TYPE="library"
+  elif (( $# == 2))
+  then
+    PROJECT_NAME="$1"
+    PROJECT_DIRECTORY="$2"
+    PROJECT_TYPE="app"
+  else
+    echo "Usage: publish_sbt_app.sh <PROJECT> [<PROJECT_DIRECTORY> | --is-library]" >&2
+    exit 1
+  fi
```
jamieparkinson, line 78: 😱 aren't bash scripts wonderful - is there a better way of doing this?

.buildkite/scripts/redeploy_ecs_services.sh
alexwlchan: No. I’ll capture some output and paste it here.

.buildkite/scripts/run_scala_task.sh
alexwlchan: Yes, I’m working on that now.

alexwlchan: I found a way to get sbt to dump the project directory, which simplifies this two ways:

1. We don't need to supply a PROJECT_DIRECTORY argument, because sbt will tell us
2. We don't need to check if this is an app/library, we just look for a Dockerfile in that directory

So now it has a much simpler interface.

.buildkite/scripts/redeploy_ecs_services.sh
alexwlchan:
```console
$ env AWS_PROFILE=catalogue-dev env CLUSTER=catalogue-api-2021-04-26 .buildkite/scripts/deploy_ecs_services.sh stage-concepts-api stage-items-api stage-search-api
Triggering a new deployment of stage-concepts-api in catalogue-api-2021-04-26
Triggering a new deployment of stage-items-api in catalogue-api-2021-04-26
Triggering a new deployment of stage-search-api in catalogue-api-2021-04-26
Waiting for stage-concepts-api to be stable
Done! stage-concepts-api is stable
Waiting for stage-items-api to be stable
Done! stage-items-api is stable
Waiting for stage-search-api to be stable
Done! stage-search-api is stable
```

jamieparkinson (approved):

PR #571, opened 2022-11-08
Expunge the last vestiges of Python from Buildkite

alexwlchan:
We're using shell scripts for the rest of Buildkite, and the Python script to run autoformatting is just a wrapper around shell commands -- let's do this everywhere. Plus bonus commentary explaining why this script exists.

I've also changed the way we run scalafmt slightly; previously we'd use whatever version was defined in the prebuilt scalafmt container, now we run sbt in the repo and use whatever version is defined there. I think this might fix our issues with mismatched scalafmt versions.

(Or maybe not? When I run the script locally it's changed a bunch of untouched files, but in CI it's done nothing.)

For https://github.com/wellcomecollection/platform/issues/5626


.buildkite/scripts/run_autoformatting.sh
```diff
@@ -0,0 +1,53 @@
...
+  git push ssh-origin "HEAD:$BUILDKITE_BRANCH"
+
+  # We exit here to fail the build, so Buildkite will skip to the next
+  # build, which includes the autoformat commit.
+  exit 1
+else
+  echo "There were no changes from auto-formatting"
+fi
+
+./builds/run_linting.sh
```
paul-butcher, line 53: run_linting.sh doesn't do auto formatting, AFAICT.  It runs flake8, and will fail if there are violations, but it doesn't autofix any.

Perhaps it belongs in its own build step

alexwlchan: I've split it out slightly; this is a vestige of some very old ways of doing things.

jamieparkinson (approved):

PR #584, opened 2022-11-17
Fix 404s in concepts API

jamieparkinson:
Previously 404s were 500ing in the concepts API (which was actually sort of useful for alerting us to the problem with missing concepts) - this fixes that, and more importantly it adds some tests

### DIFF
--- a/concepts/src/controllers/concept.ts
+++ b/concepts/src/controllers/concept.ts
@@ -1,3 +1,4 @@
+import { errors as elasticErrors } from "@elastic/elasticsearch";
 import { RequestHandler } from "express";
 import asyncHandler from "express-async-handler";
 import { Clients, Concept, Displayable } from "../types";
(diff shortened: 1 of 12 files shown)


concepts/src/controllers/concept.ts
```diff
@@ -17,21 +18,26 @@ const conceptController = (
...
+    try {
+      const getResponse = await elasticClient.get<Displayable<Concept>>({
+        index,
+        id,
+        _source: ["display"],
       });
-    }
 
-    res.status(200).json(getResponse._source.display);
+      res.status(200).json(getResponse._source!.display);
```
alexwlchan, line 28: What happens if `_source` is missing? In the old code, that would send us down the 404 path – what happens in the new code?

concepts/src/services/logging.ts
```diff
@@ -1,39 +1,22 @@
...
   format.timestamp({ format: "HH:mm:ss.SSS" }),
   format.errors({ stack: true }),
-  format.printf(
-    ({ level, message, timestamp }) =>
-      `${timestamp} ${level.toUpperCase()} - ${message}`
-  )
+  format.printf(({ level, message, timestamp, stack }) => {
+    let log = `${timestamp} ${level.toUpperCase()} - ${message}`;
+    if (stack) {
+      log += "\n" + stack;
```
alexwlchan, line 10: Given we have more fine-grained control over the logging here, is there anything we can do to make these a bit nicer in the logging cluster? e.g. using `r` instead of `n`.

concepts/src/controllers/concept.ts
jamieparkinson: It isn't possible for it to be missing if the `get` hasn't thrown, but unfortunately this isn't enforced by the types Elastic provide - it would 500, which I am fine with because it would be very much unexpected.

alexwlchan (approved):
Thanks for clarifying the type coercion question. The logging stuff isn't blocking if it's equivalent to our current Scala, just a thought while reading. Tests look good.

concepts/src/services/logging.ts
jamieparkinson: Ah yes, good idea - ideal world I would like us to switch to proper structured logging (probably https://www.elastic.co/guide/en/ecs/current/ecs-reference.html), but not right now

PR #591, opened 2022-12-07
Fixes: Test discovery / KNN / tests that broke during broken discovery

paul-butcher:
Test discovery has been broken for a while, unfortunately this has meant that

a) The new KNN stuff never ran any tests, so it never failed, but it was a bit broken
b) Other tests have been silently failing for a while, partly due to changes in test data without updated 

So, this PR is a bit scattery, as it is not only fixing test discovery, but also bringing the build back to green having done so.

1. It fixes test discovery
2. It fixes KNN, 
3. It adds new test data and updates relevant expectations
4. It reduces _some_ of the fragility relating to changes in test data


search/src/test/scala/weco/api/search/fixtures/ApiFixture.scala
```diff
@@ -83,20 +84,51 @@ trait ApiFixture extends AnyFunSpec with ScalatestRouteTest with IndexFixtures {
...
             contentType shouldEqual ContentTypes.`application/json`
-            parseJson(responseAs[String], unordered) shouldEqual parseJson(
-              expectedJson,
-              unordered
-            )
             status shouldEqual expectedStatus
+            parseJson(responseAs[String], unordered)
           }
+        }
+        responseJson shouldEqual parseJson(expectedJson, unordered)
```
paul-butcher, line 97: I have moved the "check the json" outside of the eventually.

What was happening before was that a failing test would take ages to execute, because it would exhaust all of its `eventually` retries checking whether the returned JSON matched.  This also gave a cryptic failure message about timing out, when the real problem is that the data is wrong.

What we really want is "eventually we get a response" then "the response is...".  That way, a request that successfully returns the wrong data can fail promptly and with the right reason.

search/src/test/scala/weco/api/search/fixtures/ApiFixture.scala
```diff
@@ -83,20 +84,51 @@ trait ApiFixture extends AnyFunSpec with ScalatestRouteTest with IndexFixtures {
...
+    }
+
+  def assertJsonResponseLike(
```
paul-butcher, line 100: These two new assertions are part of removing the fragility, but also make it easier to read the tests to understand what they are looking for.

Previously, the tests would all make a request, then check whether that whole response matched the content of the stored json file.

Unfortunately, this means that when we regenerate the test documents, we then had to modify the  expectations accordingly.

With this change, we can simply inspect the relevant parts of the response document.

search/src/test/scala/weco/api/search/images/ImagesSimilarityTest.scala
```diff
@@ -1,85 +1,126 @@
...
+    * for populating the similar image lists.  Essentially, this is showing the
+    * connection between the `include` querystring parameter and the resulting JSON.
+    *
+    * The thresholds are set too high to be able to actually return similar images in
+    * these dummy indices, so the content of these image lists is irrelevant to the tests.
+    *
+    * Tests that show that the similar images functionality behaves as expected
+    * can be found in ../services/ImagesServiceTest.scala
+    * and there should also be rank tests to cover whether the images returned are appropriately similar
+    */
```
alexwlchan, line 16: This is a good comment; I wonder if it would be leaning on these tests to check some of the new kNN behaviour is working correctly? i.e. that images are never similar to themselves

While I was testing locally, I commented out the code in ImagesService that filters out an image from the kNN response, but this test still passes.

So the assertions become something like "the JSON contains `visuallySimilar` _and_ it's an empty list".

alexwlchan (commented):
Overall this looks pretty good, and some of the test fragility stuff is nice. I've tried running it locally and tests pass. Thanks for getting this all sorted. 🥳 

A bunch of small suggestions inline; the big one is whether we can tweak the kNN query so that Elasticsearch filters out the original image for us, rather than us duplicating that logic/knowledge across multiple files.

search/src/main/scala/weco/api/search/services/ImagesRequestBuilder.scala
```diff
@@ -141,7 +146,30 @@ class ImagesRequestBuilder(queryConfig: QueryConfig)
...
+  private def rawSimilarityRequest(
+    query: (String, IndexedImage, Index) => JsonObject
+  )(
+    index: Index,
+    imageId: String,
+    image: IndexedImage,
+    n: Int,
+    minScore: Double
+  ): SearchRequest = {
+    val x = Json
```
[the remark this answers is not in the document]
paul-butcher: Whoops.  I didn't intend to keep this as a separate variable, it was just while I was trying to build it all up.

search/src/main/scala/weco/api/search/services/ImagesService.scala
```diff
@@ -71,8 +71,17 @@ class ImagesService(
     elasticsearchService
       .findBySearch(searchRequest)(decoder)
       .map {
-        case Left(_)       => Nil
-        case Right(images) => images
+        case Left(err) =>
+          println(err)
+          Nil // TODO: This should probably log the error
```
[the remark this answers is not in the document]
paul-butcher: toString on that error isn't very helpful, I was thinking of making it actually write out the error as returned by Elasticsearch or some interpretation of that e.g. SearchPhaseExecutionError.reason.

However, that's a bit out of scope for this change. I'll just replace it back to Nil for now and put in a separate change for it

search/src/main/scala/weco/api/search/services/ImagesService.scala
```diff
@@ -71,8 +71,17 @@ class ImagesService(
...
+        case Right(images) =>
+          // The response from a KNN query includes the image we were searching _with_
+          // This is because, unlike MLT queries, the client must provide the data to match,
+          // rather than asking ES to find things similar to a record it already knows about.
+          // So, of course, the record that has a 100% match will be included in the response.
```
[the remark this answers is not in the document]
paul-butcher: That is certainly the way forward.  You may be able to tell that this was the point at which I was getting a bit muttery on Tuesday.  I had a baffling moment where this switched from not working at all to almost working, followed by the epiphany that made me write this bit.

search/src/test/scala/weco/api/search/images/ImagesSimilarityTest.scala
paul-butcher: I think that's a good plan.  These tests originally loaded half a dozen similar records, then checked whether the list for one contained all the others.  When the thresholds changed some months ago, this was still loading them all but now checking for an empty list.

I thought that would be fragile, as the random fuzz in two similar records could possibly result in documents similar enough to get into the list.

Then I removed all the redundant source records, but didn't make the leap that this meant the test could now rely on the lists to be empty.

builds/run_sbt_tests.sh
```diff
@@ -34,7 +34,7 @@ echo "*** Running sbt tests"
 
 if [[ -f "$PROJECT_DIRECTORY/docker-compose.yml" ]]
 then
-  ./builds/run_sbt_task_in_docker.sh "project $PROJECT_NAME" "dockerComposeTest"
```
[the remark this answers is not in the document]
paul-butcher: I think we should.  I don't know what it is about this command that breaks discovery, because it does find some tests, just not all of them.   If these other projects have tests with whatever characteristic means they can't be found by dockerComposeTest, then we could have quite a widespread problem.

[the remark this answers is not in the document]
alexwlchan: 🤞 

https://github.com/wellcomecollection/scala-libs/pull/197
https://github.com/wellcomecollection/storage-service/pull/1038
https://github.com/wellcomecollection/catalogue-api/pull/593

search/src/main/scala/weco/api/search/elasticsearch/ImagesQuery.scala
```diff
@@ -49,23 +48,32 @@ case object ImageSimilarity {
...
+          "num_candidates" -> 100.asJson,
+          "filter" -> Json.obj(
+            "bool" -> Json.obj(
+              "must_not" -> Json.obj(
+                "ids" -> Json.obj(
+                  "values" -> Json.arr(Json.fromString(imageId))
+                )
+              )
+            )
+          )
```
alexwlchan, line 72: 🥳 

The amount of `Json.method` we're having to sprinkle around is a bit icky, but that's a problem for another day.

search/src/main/scala/weco/api/search/services/ImagesRequestBuilder.scala
```diff
@@ -141,7 +147,30 @@ class ImagesRequestBuilder(queryConfig: QueryConfig)
...
+    search(index).source(
+      Json
+        .fromJsonObject(
+          query(imageId, image, index)
+            .add("min_score", Json.fromDouble(minScore).get)
+            // n is the number of similar documents we want the query to return
+            // However, because this is used by KNN searches, which also return
+            // the document in question (which is later filtered out of the result list),
+            // an extra slot must be added for the source document to occupy.
+            .add("size", Json.fromInt(n + 1))
```
alexwlchan, line 172: Do we still need to fetch `n + 1` images if we're excluding the original image at source?

search/src/main/scala/weco/api/search/services/ImagesService.scala
```diff
@@ -55,11 +55,12 @@ class ImagesService(
         requestBuilder.requestWithSimilarColors
     }
 
-    // default minimum scores for each similarity metric determined using this notebook
+    // default minimum score for the colors similarity metric determined using this notebook
     // https://github.com/wellcomecollection/data-science/blob/47245826c70bf2d76c63d2c4b3ace6c824673784/notebooks/similarity_problems/notebooks/01-similarity-scores.ipynb
+    // The blended and features metric use KNN which gives a value between 0 and 1.  The ideal threshold value is yet to be determined.
```
alexwlchan, line 60: Good comment. 👍

search/src/test/scala/weco/api/search/fixtures/ApiFixture.scala
```diff
@@ -83,20 +84,50 @@ trait ApiFixture extends AnyFunSpec with ScalatestRouteTest with IndexFixtures {
...
+      Get(path) ~> routes ~> check {
+        status shouldEqual Status.OK
+        contentType shouldEqual ContentTypes.`application/json`
+        parseJson(responseAs[String])
       }
     }
+    locator(responseJson) shouldEqual parseJson(
+      expectedJson
+    )
+  }
```
alexwlchan, line 130: minor suggestion: can you define this in terms of the previous fixture?

```suggestion
  def assertJsonResponseContains(
    routes: Route,
    path: String,
    locator: Json => Json,
    expectedJson: String
  ): Assertion =
    assertJsonResponseLike(routes, path) { responseJson =>
      locator(responseJson) shouldEqual parseJson(
        expectedJson
      )
    }
```

alexwlchan (approved):
Looks great, a couple of tiny comments but happy for it to go in as-is if that's easier. 🖌️

PR #595, opened 2023-01-03
Add additional publication fields for serials

alexwlchan:
* The `currentFrequency` field is an optional string and so will always come through. The API doesn't need to change.
* The `formerFrequency` and `designation` fields are lists, so we want them to be include-able to match our other API parameters.

For https://github.com/wellcomecollection/platform/issues/5600


jamieparkinson (approved):
✨

PR #618, opened 2023-02-08
Snapshot S3 bucket updates

jamieparkinson:
- Fix deprecation warnings
- Add a CORS policy for https://github.com/wellcomecollection/developers.wellcomecollection.org/pull/17


terraform/snapshots/domain.tf
```diff
@@ -15,21 +15,33 @@ resource "aws_acm_certificate" "data_page" {
   }
 }
 
+moved {
+  from = aws_route53_record.cert_validation
+  to   = aws_route53_record.cert_validation["data.wellcomecollection.org"]
+}
```
alexwlchan, line 21: minor: it's good to have these in the commit history, but I prefer to delete them once applied so they don't clutter up the codebase.

Doing this with a `for_each` is nice. 👍

alexwlchan (approved):

terraform/snapshots/domain.tf
jamieparkinson: Oops, forgot it was here - tf actually forced me to do the `for_each` as it correctly points out that a set, being unordered, doesn't have a `0`th element.

PR #619, opened 2023-02-14
Rank test order

harrisonpim:
- adds the ability to explicitly test the order of results in rank
- adds some tests for cases where we care about the order of results
- updates the query to use new fields from https://github.com/wellcomecollection/catalogue-pipeline/pull/2329

closes https://github.com/wellcomecollection/catalogue-api/issues/607 and closes https://github.com/wellcomecollection/catalogue-api/issues/608


docs/search/current_queries/works.md
```diff
@@ -7,109 +7,155 @@
...
-          "type": "best_fields",
-          "analyzer": "whitespace_analyzer",
-          "operator": "Or",
-          "_name": "identifiers"
-        }
-      },
-      {
-        "dis_max": {
-          "queries": [
+    "bool": {
```
alexwlchan, line 10: Could we just link to a file here, to save us having to keep this up-to-date?

rank/scripts/test.ts
```diff
@@ -0,0 +1,62 @@
...
+  testIds.map(async (testId) => {
+    const result =
+      testId === 'order'
+        ? await testOrder({ queryEnv, index, testId, cluster })
+        : await rankTest({ queryEnv, index, testId, cluster })
+
+    return result.results.map((result) => {
+      if (result.knownFailure) {
+        return {
+          message: `🟡\t"${result.searchTerms}" is a known failure\n\t${result.description}\n`,
```
alexwlchan, line 34:
```suggestion
          message: `⚠️\t"${result.searchTerms}" is a known failure\n\t${result.description}\n`,
```

rank/scripts/test.ts
```diff
@@ -0,0 +1,62 @@
...
+          pass: true
+        }
+      }
+      if (result.result.pass) {
+        return {
+          message: `🟢\t"${result.searchTerms}" passes\n\t${result.description}\n`,
```
alexwlchan, line 40:
```suggestion
          message: `✅\t"${result.searchTerms}" passes\n\t${result.description}\n`,
```

rank/scripts/test.ts
```diff
@@ -0,0 +1,62 @@
...
+          pass: true
+        }
+      } else {
+        return {
+          message: `🔴\t"${result.searchTerms}" fails but should pass\n\t${result.description}\n`,
```
alexwlchan, line 45:
```suggestion
          message: `❌\t"${result.searchTerms}" fails but should pass\n\t${result.description}\n`,
```

distinguishing by colour and shape > distinguishing by colour alone

rank/src/services/order.ts
```diff
@@ -0,0 +1,79 @@
...
+  const template = await getTemplate(queryEnv, index)
+
+  const test = tests[template.namespace].find((test) => test.id === testId)
+  if (!test) throw Error(`No such test ${testId}`)
+
+  const results = await Promise.all(
+    // for each test case, run a search
+    test.cases.map(async (testCase) => {
+      const { searchTerms, description, knownFailure } = testCase
+      const client = await getClient(queryEnv, cluster, template)
```
alexwlchan, line 41: Can you move this out to the top-level, reuse the same client rather than fetching it each time?

alexwlchan (commented):
I think I need at least one more pass to wrap my head around this, but looks okay so far.

.buildkite/scripts/test_rank.sh
```diff
@@ -12,11 +12,11 @@ docker run \
...
 # Copy candidate queries into place.
-# TODO: This is a short-term fix to unbreak the build, find a better
-# way to do this.
-ROOT=$(git rev-parse --show-toplevel)
-cp "$ROOT/search/src/test/resources/WorksMultiMatcherQuery.json" "$ROOT/rank/queries/WorksMultiMatcherQuery.json"
-cp "$ROOT/search/src/test/resources/ImagesMultiMatcherQuery.json" "$ROOT/rank/queries/ImagesMultiMatcherQuery.json"
+docker run \
+    --volume $(pwd):/catalogue-api \
+    --workdir /catalogue-api/rank \
+    public.ecr.aws/docker/library/node:14-slim \
```
alexwlchan, line 18: v minor: we've been moving to Node 16 or Node 18 in the front-end repos. Not a big deal, but let's make a note to move this up at some point.

docs/search/current_queries/works.md
```diff
@@ -2,117 +2,7 @@
...
-        }
-      }
-    ],
-    "minimum_should_match": "1"
-  }
-}
-```
-
-</details>
+[click here for the complete query json](../../../rank/data/queries/WorksMultiMatcherQuery.json)
```
alexwlchan, line 5: sweet

look at all that red!

docs/search/current_queries/works.md
```diff
@@ -275,3 +165,23 @@ we want to prioritise the third result. Based on user feedback documented here <
...
+            "query.title.shingles_cased^1000.0",
+            "query.alternativeTitles.shingles_cased^100.0",
+            "query.partOf.title.shingles_cased^10.0"
+        ],
+        "type": "most_fields",
+        "operator": "And",
+        "_name": "shingles_cased"
+    }
+}
+```
```
alexwlchan, line 187: Not sure if this was here on the first pass, but this is really nice (and potentially solves the problem we discussed about comments in the mapping/query).

This feels like it could be edited into a document we could share with non-data scientists, e.g. Collections or me – each section starts with a digestible summary and a few examples, and then you hide the "how it works" behind a collapsible `<details>` block.

That's not a "now" problem, but more a thought for the future.

rank/data/tests/images.json
```diff
@@ -0,0 +1,142 @@
...
+    {
+        "id": "precision",
+        "description": "TBD",
+        "eval": "equalTo1",
+        "cases": [
+            {
+                "searchTerms": "crick dna sketch",
+                "ratings": [
+                    "gzv2hhgy"
+                ]
```
alexwlchan, line 11: Another "not for now" problem: I wonder if we could change this data structure slightly to allow a comment alongside these IDs?

I have to look them up in /works to see what they are, which is fine right now, but could cause issues if e.g. IDs get suppressed or redirected – a brief comment about the intention might be helpful.

```
"ratings": [
    {
        "id": "gzv2hhgy",
        "comment": "Pencil sketch of the DNA double helix by Francis Crick"
    }
]
```

rank/data/tests/images.json
```diff
@@ -0,0 +1,142 @@
...
+        "id": "alternative-spellings",
+        "description": "TBD",
+        "eval": "equalTo1",
+        "cases": [
+            {
+                "searchTerms": "arbeiten",
+                "ratings": [
+                    "sr4kxmk3",
+                    "utbtee43"
+                ]
```
alexwlchan, line 118: Similar to above, and esp because metadata changes – it'd be good to know what you thought the similar text was in the image.

It's not clear to me what it is in either case – `sr4kxmk3` literally contains "Arbeiten" in the title which is just cased differently, and I have no idea what's being matched on `utbtee43`.

rank/data/tests/works.json
```diff
@@ -0,0 +1,439 @@
...
+                ],
+                "description": "Reference number as ID"
+            },
+            {
+                "searchTerms": "2599i",
+                "ratings": [
+                    "xxskepr5"
+                ],
+                "description": "Reference number as ID"
+            },
```
alexwlchan, line 34: Maybe toss in a CALM reference as a search as well?

rank/data/tests/works.json
```diff
@@ -0,0 +1,439 @@
...
+            {
+                "searchTerms": "WA/HMM/CM benin",
+                "ratings": [
+                    "qfdvkegw",
+                    "je5pm2gj",
+                    "dppjjtqz"
+                ],
+                "description": "Archive refno and a word from the description"
+            },
```
alexwlchan, line 179: Extreme nitpick: what you're testing is the Calm <strong>Alt</strong>RefNo, not the RefNo.

The RefNo is the absolutely structured thing we use to create the archive trees, which in this case would be `WAHMM/CM`, and should never change. The AltRefNo is the publicly visible thing, `WA/HMM/CM`, which can sometimes change. It's for display purposes only.

We have two because the first prefix before a slash serves two different purposes:

* In the RefNo it identifies everything inside this archive; everything matching `WAHMM/*` is part of the Wellcome Historical Medical Museum. Everything starting `WA/*` would be too broad to be useful.
* In the AltRefNo it tells the reader what sort of archive this is, e.g. `PP/*` is personal papers (I don't remember what `WA/*` is)

I would suggest:

1. Change this description to say `archive reference`
2. Add a test for the RefNo, which is `WAHMM/CM`. Probably only used by staff, but should still work in results.

rank/scripts/getIndexConfig.ts
```diff
@@ -19,13 +19,11 @@ async function go() {
...
-    const mappingsReq = client.indices.getMapping({ index: indices })
-    const [{ body: settingsRes }, { body: mappingsRes }] = await Promise.all([
-      settingsReq,
-      mappingsReq,
-    ])
+    const settingsRes = await client.indices.getSettings({ index: indices })
 
+    const mappingsRes = await client.indices.getMapping({ index: indices })
+
+    console.log(mappingsRes)
```
alexwlchan, line 26: left over from debugging?

rank/src/services/order.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  const afterHits = searchResponse.hits.hits.filter((hit) =>
+    testCase.after.includes(hit._id)
+  )
+
+  const pass = beforeHits.every((beforeHit) => {
+    return afterHits.every((afterHit) => {
+      return beforeHit._score > afterHit._score
+    })
+  })
```
alexwlchan, line 20: This is fine because `beforeHits` and `afterHits` are small, but it's O(N * M) in the size of the two lists which could get thorny if you had large test cases.

Another way to structure this would be:

```typescript
const afterScores = afterHits.map((hit) => hit._score);
const beforeScores = beforeHits.map((hit) => hit._score);

return Math.min(afterScores) > Math.max(beforeScores);
```

which is only O(N + M) because you iterate through each list twice.

I don't think it's a big deal and you don't need to change it here – the majority of the rank runtime will be spent waiting for network requests – but I note for interest.

rank/src/services/test.ts
```diff
@@ -27,73 +25,65 @@ const indexExists = async (client: Client, index: string): Promise<boolean> => {
   }
 }
 
-async function service({
+export async function getClient(queryEnv, cluster, template): Promise<Client> {
+  if (queryEnv === 'candidate' && cluster === 'pipeline') {
```
alexwlchan, line 29: nit: flip the order of conditions, so it's easier to see this is a special instance of the next case

```suggestion
  if (cluster === 'pipeline' && queryEnv === 'candidate') {
```

alexwlchan (approved):
Some suggestions about how we document the rank test cases which you could consider for another patch, but nothing that should block (or add to!) this particular PR.

PR #621, opened 2023-02-21
Handle the case where no updates were made

jamieparkinson:
The logic for how we found the last update wasn't quite right, this fixes that

### DIFF
--- a/snapshots/snapshot_reporter/src/snapshot_reporter.py
+++ b/snapshots/snapshot_reporter/src/snapshot_reporter.py
@@ -277,12 +277,12 @@ def get_recent_update_stats(session, *, hours):
         "works": {
             "index": works_index_name,
             "count": works_count,
-            "last_update": works_last_update
+            "last_update": parser.parse(works_last_update),
         },
         "images": {
             "index": images_index_name,
             "count": images_count,
-            "last_update": images_last_update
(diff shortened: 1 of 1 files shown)


snapshots/snapshot_reporter/src/snapshot_reporter.py
```diff
@@ -252,30 +253,38 @@ def get_recent_update_stats(session, *, hours):
...
+    images_count = pipeline_es_client.count(
+        index=images_index_name, body=indexed_time_body
+    )["count"]
+
+    works_last_update = pipeline_es_client.search(
+        index=works_index_name, body=last_update_body
+    )["hits"]["hits"][0]["_source"]["debug"]["indexedTime"]
+    images_last_update = pipeline_es_client.search(
+        index=images_index_name, body=last_update_body
+    )["hits"]["hits"][0]["_source"]["debug"]["indexedTime"]
```
alexwlchan, line 275: This is the sort of place where I really miss either (1) JavaScript’s dot syntax for looking up properties or (2) the ability to do multi-line commands without breaking out backslash escapes – I find heavily nested dictionaries kinda ugly in Python.

(Nothing you can do about it, just musing!)

alexwlchan (approved):
🕙

PR #625, opened 2023-03-01
Query with minimum should match

harrisonpim:
This PR:
- adds new rank tests based on feedback in https://github.com/wellcomecollection/catalogue-api/issues/623
- changes all query ANDs to ORs with a `minimum_should_match` set at `80%` 

Still need to update the docs to describe the changes we've made and why

Tests will probably fail until https://github.com/wellcomecollection/catalogue-pipeline/pull/2334 has been merged and deployed, but the candidate query and mapping produce the following results:

<details>

```
➜ AWS_PROFILE=platform-dev yarn rank
yarn run v1.22.19
$ ts-node ./scripts/rank
✔ cluster › rank
✔ query › candidate
✔ index › works-candidate
Running:
  yarn test --index=works-candidate --queryEnv=candidate --cluster=rank --testId=precision --testId=recall --testId=alternative-spellings --testId=order

$ ts-node ./scripts/test --index=works-candidate --queryEnv=candidate --cluster=rank --testId=precision --testId=recall --testId=alternative-spellings --testId=order
✅      "stimming" passes
        Ensure that we return non-typos over typos e.g. query:stimming matches:stimming > swimming

⚠️       "The Piggle" is a known failure
        Example of a known title's prefix, but not the full thing

✅      "DJmjW2cU" passes
        Case insensitive IDs

✅      "Bills of mortality" passes
        Example of a known title's prefix, but not the full thing

✅      "sophies shell" passes
        Searching without punctuation should match a document with punctuation

✅      "Das neue Naturheilverfahren" passes
        Example of a known title's prefix, but not the full thing

✅      "Oxford dictionary of national biography" passes
        Example of a known title's prefix, but not the full thing

✅      "stim" passes
        Exact match on title with lowercasing and punctuation stripping

✅      "bulloch history of bacteriology" passes
        Contributor and title in the same query

✅      "L0033046" passes
        Miro ID matching

✅      "kmebmktz" passes
        Work ID matching

✅      "mammas favorites" passes
        Searching without punctuation should match a document with punctuation

✅      "seq88sr4 qfk4vbp8" passes
        multiple IDs

✅      "Cassils Time lapse" passes
        Contributor and title in the same query

✅      "information law" passes
        Multi-word exact matches at the start of a title should be prioritised https://github.com/wellcomecollection/catalogue-api/issues/466

✅      "2013i" passes
        Reference number as ID

✅      "2599i" passes
        Reference number as ID

✅      "gzv2hhgy" passes
        Image ID matching

[...]

PR #628, opened 2023-03-02
Allow filters in rank

### LINKED ISSUE #626: rank should be able to run tests with filters applied
[see thread here](https://wellcome.slack.com/archives/C8X9YKM5X/p1677748509274129)

To test for these cases, we need to be able to apply filters to individual rank queries

harrisonpim:
Adds the ability to filter results when building rank tests, or when searching through the CLI

closes #626 

Rank tests won't pass until https://github.com/wellcomecollection/catalogue-api/pull/625 and https://github.com/wellcomecollection/catalogue-pipeline/pull/2334 have been merged and deployed.

### DIFF
--- a/rank/data/tests/works.json
+++ b/rank/data/tests/works.json
@@ -5,6 +5,18 @@
         "eval": "equalTo1",
         "cases": [
             {
+                "searchTerms": "Mackie",
+                "ratings": [
+                    "k8ccmj6t"
+                ],
+                "description": "An exact match in the title and description, and filtering for archives",
+                "filter": {
+                    "term": {
+                        "query.format.id": "h"
+                    }
+                }
+            },
+            {
                 "searchTerms": "information law",
                 "ratings": [
                     "zkg7xqm7"
@@ -452,7 +464,11 @@
                     "ae5kymcs"
                 ],
                 "description": "An exact matches in the title should appear before a misspelled match in the contributor and title. Also uses a filter",
-                "filter": {"term": {"query.format.id": "h"}}
+                "filter": {
+                    "term": {
+                        "query.format.id": "h"
+                    }
+                }
             }
         ]
     }
--- a/rank/data/tests/works.json
+++ b/rank/data/tests/works.json
@@ -119,6 +119,14 @@
                 ],
                 "description": "Example of a known title's prefix, but not the full thing",
                 "knownFailure": true
+            },
+            {
+                "searchTerms": "Hunterian wa/hmm",
+                "ratings": [
+                    "f3gpbk74"
+                ],
+                "description": "archive reference number and a word from the title",
+                "knownFailure": true
             }
         ],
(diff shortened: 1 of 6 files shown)


rank/data/tests/works.json
```diff
@@ -433,6 +441,18 @@
...
+            {
+                "searchTerms": "Mackie",
+                "before": [
+                    "k8ccmj6t",
+                    "s2tnttyn"
+                ],
+                "after": [
+                    "ae5kymcs"
+                ],
+                "description": "An exact matches in the title should appear before a misspelled match in the contributor and title. Also uses a filter",
```
alexwlchan, line 454: Since the filter is just a boolean yes/no, it doesn't affect the order of results.

If we want to test the filter logic is working, can we add an example of something that _shouldn't_ appear because it's being excluded by the filter?

harrisonpim: Maybe... IMO this set of tests should only exist to validate relevance, not the rank tool itself, and I'm confident that the filters are being applied when a query is run. I think more and more of our tests will start incorporating filters as time goes on, but I'm hesitant to add tests prematurely which aren't motivated by real user needs!

Maybe we can include an additional `Mackie` example in the `precision` tests to really incorporate the filter feature, ie

<details><summary>Searching for Mackie without a filter</summary>

```
✔ index › works-candidate
✔ queryEnv › candidate
✔ What are you looking for? … Mackie
✔ Any filters? … 
1.      Jog on : how running saved my life / Bella Mackie.
        [2018]
        https://wellcomecollection.org/works/kqczv6ks

2.      Tapping the deep roots of Britain's Blacks / Lindsay Mackie.
        1980
        https://wellcomecollection.org/works/vpuqcn4c

3.      Notes of a case of persistent priapism / by John Mackie.
        1872
        https://wellcomecollection.org/works/vdzp9u3d
```

</details>

<details><summary>Searching for Mackie with a filter</summary>

```
✔ index › works-candidate
✔ queryEnv › candidate
✔ What are you looking for? … Mackie
✔ Any filters? … {"term": {"query.format.id": "h"}}
1.      Mackie, Donald
        HALDANE/5/2/3/52
        26 Dec 1952
        https://wellcomecollection.org/works/k8ccmj6t

2.      Mackie J
        WF/M/I/PR/727b
        c.1984
        https://wellcomecollection.org/works/s2tnttyn

3.      ACL Mackie
        SA/ASH/C/18
        1972-1973
        https://wellcomecollection.org/works/y3ghxj83
```

</details>

So our test should ensure that `k8ccmj6t` is first in the list when an archive filter is applied

alexwlchan: Mmm, okay. That makes sense. (I'm now reading the Slack thread.)

Let's leave this as-is then.

alexwlchan (approved):

PR #631, opened 2023-03-14
Customise the akka-http timeout response to be JSON

alexwlchan:
The front-end expects that the catalogue API will always return errors in JSON, but if we get a timeout we're returning the default akka-http response:

    The server was not able to produce a timely response to your request.
    Please try again in a short while!⏎

This wraps that error in our JSON error type, which should make it easier for the front-end to handle.

I considered adding a test for this, but this particular query doesn't timeout with an empty index (unsurprisingly) and I don't think it's worth the time to get a full repro.

For https://github.com/wellcomecollection/platform/issues/5665


paul-butcher (approved):
⏳

PR #639, opened 2023-03-29
Genre concept ids

paul-butcher:
This adds the ability to filter on genre concepts by id, following https://github.com/wellcomecollection/catalogue-pipeline/pull/2344

Swagger update: https://github.com/wellcomecollection/developers.wellcomecollection.org/pull/19

https://github.com/wellcomecollection/catalogue-api/issues/637

New issue to extend the approach for empty filters across the API: 
https://github.com/wellcomecollection/catalogue-api/issues/647


search/src/test/scala/weco/api/search/works/WorksIncludesTest.scala
paul-butcher, line 1: Because I changed something a bit fundamental in the output, the randomiser ended up causing quite a bit of churn in this file, and on the list tests in particular, I found it hard to see what the important part of the test is. Fortunately, there was nothing in the application that would have changed the output, so it was easy enough to check and be confident that all the changes in actual output were down to the changes in the input data.

I have raised a refactoring ticket that might help https://github.com/wellcomecollection/catalogue-api/issues/640. I had started implementing those changes as part of this, but realised that it was going to be a significant change leading to an even harder review.

search/docker-compose.yml
```diff
@@ -13,3 +13,4 @@ services:
       - "logger.level=DEBUG"
       - "discovery.type=single-node"
       - "xpack.security.enabled=false"
+      - "cluster.routing.allocation.disk.threshold_enabled=false"
```
alexwlchan, line 16: What does this do?

search/src/main/scala/weco/api/search/rest/ImagesParams.scala
```diff
@@ -42,6 +42,7 @@ case class MultipleImagesParams(
   license: Option[LicenseFilter],
   `source.contributors.agent.label`: Option[ContributorsFilter],
   `source.genres.label`: Option[GenreFilter],
+  `source.genres.concepts.id`: Option[GenreConceptFilter],
```
alexwlchan, line 45: When the final part of a filter is the `.id`, we omit the `.id` – this filter should be called `source.genres.concepts`, and similarly for `genre.concepts` in the catalogue API.

See https://github.com/wellcomecollection/docs/tree/main/rfcs/037-api-faceting-principles#principles, which we should probably put somewhere more prominent.

search/src/main/scala/weco/api/search/services/ImagesRequestBuilder.scala
```diff
@@ -112,12 +111,15 @@ class ImagesRequestBuilder(queryConfig: QueryConfig)
         )
       case GenreFilter(genreLabels) =>
         termsQuery("query.source.genres.label.keyword", genreLabels)
+      case GenreConceptFilter(conceptIds) =>
+        if (conceptIds.isEmpty) NoopQuery
+        else termsQuery("query.source.genres.concepts.id", conceptIds)
```
alexwlchan, line 116: Do we need a special case for no concept IDs passed for genre filters only, or do you think we should be doing this for all filters?

alexwlchan (commented):
The tests and code look good, just a note about API naming conventions and handling the case of no genres passed.

search/src/main/scala/weco/api/search/services/ImagesRequestBuilder.scala
paul-butcher: I think it should be for all filters.  The current behaviour for other filters is that an empty filter causes it not to match anything, which I discovered when adding the Cardinality test for this.

I didn't want to extend it as part of this work, as it was already big enough.
I'll add a ticket to extend this behaviour.

search/src/main/scala/weco/api/search/rest/ImagesParams.scala
paul-butcher: Good point. Now you mention it, I remember this from ages ago.

search/docker-compose.yml
paul-butcher: When testing, I found that ES would often get into a red state because it passed the high watermark for disk space in the docker container.  This switches off that check so that it will run even if the "disk" is 95% full.

alexwlchan (approved):
🧞

PR #648, opened 2023-04-20
Add filtering and sorting by production date on images

alexwlchan:
This is the API side of https://github.com/wellcomecollection/platform/issues/5668

It turns out the correct fields are already indexed, so combining this with test documents in https://github.com/wellcomecollection/catalogue-pipeline/pull/2353, this can be deployed with the current index to get the new behaviour.


jamieparkinson (approved):
lovely

PR #652, opened 2023-05-10
update the rank tests to use a changed work id

harrisonpim:
Rank tests are failing because [djmjw2cu](https://wellcomecollection.org/works/djmjw2cu) has become [uemqqmb9](https://wellcomecollection.org/works/uemqqmb9). This PR updates the rank tests to use the new ID


alexwlchan (approved):

harrisonpim:
I'm not sure why that last referenceNumber test is still failing.

Whatever the reason, these issues should be wiped out by the upcoming changes to the mapping + query. I'm going to mark it as a `knownFailure` for the time being.

Issue #671, opened 2023-06-29
Only return non-empty (or filtered on) buckets in aggregations


paul-butcher:
https://github.com/wellcomecollection/catalogue-api/pull/674 fixes this

Issue #666, opened 2023-06-16
Improve test suite for faceting


paul-butcher:
In part, inspired by this: https://wellcome.slack.com/archives/C3TQSF63C/p1686661286133629

paul-butcher:
I've written out some Gherkin.  There are some other rules in  [Faceting Principles](https://github.com/wellcomecollection/docs/tree/main/rfcs/037-api-faceting-principles) that I'm yet to include (the one about label vs id), but I think these are the critical ones I want to cover right now.

I have so far made this applicable to any API, so the ruleset (and possibly the actual Scala features file) could be reused elsewhere. 
 
```feature
 Feature: Faceting - aggregations and filters
   Background: 
     Given a dataset with aggregable values

   Rule: Aggregations are only returned when requested 
     Aggregations are returned in the aggregations collection, 
     they can be slow and expensive, and shouldn't happen if the client doesn't want them 
       Scenario: a request with no aggregations
         When records are requested
         Then no aggregations are returned
       Scenario: a request with one aggregation
         When records are requested with an aggregation
         Then that aggregation will be returned
       Scenario: a request with multiple aggregations
         When records are requested with two aggregations
         Then both aggregations will be returned
    
   Rule: Queries constrain the document collection over which aggregations operate 
       Scenario: a request with a query and aggregations
         When a query is made with two aggregations
         Then only documents that match the query are counted in the aggregations
    
   Rule: Filters apply to all aggregated fields apart from their own 
       Scenario: filtering one one field, and aggregating on another
         filters constrain the document collection for other aggregations
         When records are requested
         And the request is filtered on one field
         And asks for aggregation on another field
         Then only documents that match the filter are counted in the aggregation
       Scenario: filtering and aggregating on the same field
         filters do not constrain the document collection when aggregating on that same field
         When records are requested
         And the request has a filter
         And asks for aggregation on the same field
         Then documents that do not match that filter are also counted in the aggregation
       Scenario: applying multiple filters to the same field
         When records are requested
         And the request has two filters on the same field
         And asks for an aggregation on the same field
         Then documents that do not match either filter are also counted in the aggregation
       Scenario: multiple filters and aggregations
         filters apply to all aggregations but their own
         When records are requested
         And the request has filters on two different fields
         And asks for an aggregation on the same fields
         And asks for an aggregation on yet another field
         Then only documents that match the filters on other fields are counted in the aggregation on each field

[...]

PR #674, opened 2023-07-06
Do not return unwanted empty buckets

paul-butcher:
In accordance with our faceting principles, empty buckets are only returned when a corresponding filter term results in an empty bucket. (https://github.com/wellcomecollection/catalogue-api/issues/671)

We have a way to do this in the Elasticsearch query as well (not yet implemented here, recently added to the Content API).

This change also adds consistent testing across the Works aggregations with paired filters.

I will be extending this consistency to Images in another ticket as part of https://github.com/wellcomecollection/catalogue-api/issues/666


search/src/main/scala/weco/api/search/models/DocumentFilterBase.scala
```diff
@@ -1,10 +1,10 @@
 package weco.api.search.models
 
 import java.time.LocalDate
+sealed trait DocumentFilterBase
```
jamieparkinson, line 4: xtreme nit: prefer `DocumentFilter` for traits so we can pretend it's not OO

search/src/test/scala/weco/api/search/works/filtering/AggregatingTestCases.scala
```diff
@@ -0,0 +1,81 @@
...
+            routes,
+            path = s"$rootPath/works?$allValuesParams"
+          ) {
+            Status.OK -> allValuesResponse
+          }
+      }
+    }
+
+    it(
+      "does not return an aggregation containing the filtered value if the value is bogus"
```
jamieparkinson, line 38: I think it would be helpful (in the variable name as well) to have a bit more context for the meaning of "bogus"

jamieparkinson (approved):
Very nice indeed, a definite improvement

Issue #668, opened 2023-06-27
Paired filter/aggregations do not work properly


paul-butcher:
When searching the catalogue on the website, selecting a (subject or contributor) [filter on a query ](https://wellcomecollection.org/search/works?query=banana&subjects.label=%22Tropical+plants%22) makes the options in that filter go awry.

Slack discussion here: https://wellcome.slack.com/archives/C02ANCYL90E/p1687857127263849

Specifically:

> Given a paired aggregation, we run a completely unfiltered terms aggregation, then we filter the results of that into a sub aggregation.
>
> So the aggregation is asking “given the top 20 subjects in the whole index, how many documents that match the query are there in each subject?”
>
> Rather than “given all the documents that match the query, what are the top 20 subjects in them, and how many documents are there in each?”
>
> This explains why the values are different for different queries, but there is also a substantial crossover.

paul-butcher:
This is a simplified representation of the kind of query that is currently happening:

```
{
  "size": 0,
  "query": {
    "match": {"query.title": "banana"}
  },
  "aggs": {
    "subjects": {
      "global": { },
      "aggs": {
        "subjects":{
        "terms": {
          "field": "query.subjects.label.keyword",
         "size": 20},
         "aggs": {
            "filtered": {
              "filter": {
                "bool": {
                  "filter": [
                    {
                      "bool": {
                        "should": [
                          {
                      "match": {"query.title": "banana"}
                    }]
                      }
                    }
                    
                  ]
                }
              }
            }
          }
          
        }
      }
      }
    } 
  }
```

which yields buckets like these:
```
         {
            "key": "Sanitation",
            "doc_count": 66847,
            "filtered": {
              "doc_count": 0
            }
          },
           {
            "key": "Germany",
            "doc_count": 4638,
            "filtered": {
              "doc_count": 1
            }
          },

**Comment by paul-butcher** [comment]
What I think is meant is a query like this:
```
{
  "size": 0,
  "query": {
    "match": {"query.title": "banana"}
  },
  "aggs": {
    "subjects": {
      "global": { },
      "aggs": {
        "subjects":{
          "filter": {
              "match": {"query.title": "banana"}
          },
          "aggs": {
            "subjects":{
            "terms": {
              "field": "query.subjects.label.keyword",
              "size": 20
           }}
          }
         }
      }
      }
    } 
  }```
which yields these as the top three buckets:

```
           {
              "key": "Bananas",
              "doc_count": 13
            },
            {
[...]

PR #676, opened 2023-07-17
Fix paired aggregation behaviour

paul-butcher:
See: [Paired filter/aggregations](https://github.com/wellcomecollection/catalogue-api/issues/668)


search/src/test/scala/weco/api/search/works/filtering/FilterAndAggregateByContributorTest.scala
```diff
@@ -59,8 +59,8 @@ class FilterAndAggregateByContributorTest
       "contributors.agent.label" -> Seq(
         (2, "Bath, Patricia"),
         (2, "Darwin \\\"Jones\\\", Charles"),
-        (1, "Jake Paul"),
-        (1, "Karl Marx")
+        (1, "Karl Marx"),
```
paul-butcher, line 62: Now that order is enforced, Karl comes before Jake

search/src/test/scala/weco/api/search/services/AggregationsTest.scala
```diff
@@ -148,35 +148,19 @@ class AggregationsTest
             List(WorkAggregationRequest.Format, WorkAggregationRequest.Subject),
           filters = List(
             FormatFilter(List("a")),
-            SubjectLabelFilter(Seq("pGkJTZWwn4"))
```
paul-butcher, line 151: This subject does not appear to exist in the test dataset. I'm a bit surprised that this test has been passing.  I suspect it's something to do with returning 0-count buckets.

search/src/test/scala/weco/api/search/services/AggregationsTest.scala
```diff
@@ -148,35 +148,19 @@ class AggregationsTest
...
+            SubjectLabelFilter(Seq("9SceRNaTEl"))
           )
         )
         whenReady(aggregationQuery(index, searchOptions)) { aggs =>
           val buckets = aggs.format.get.buckets
-          buckets.length shouldBe works.length
+          buckets.length shouldBe 7
```
paul-butcher, line 156: There was no subject in the existing dataset that was present in all works.  However, this is an improvement in that it shows the filter being applied.

search/src/main/scala/weco/api/search/services/FiltersAndAggregationsBuilder.scala
```diff
@@ -34,34 +36,61 @@ trait FiltersAndAggregationsBuilder[Filter, AggregationRequest] {
...
+        val filterTerm = filterQuery match {
+          case TermsQuery(_, values, _, _, _, _, _) =>
+            values.map(value => s"""\\"$value\\"""").mkString("|")
+          case _ => ""
+        }
+        terms.minDocCount(0).includeRegex(s".*($filterTerm).*")
+      case agg => agg
+    }
+
+  def requestToOrderedAggregation(
```
paul-butcher, line 56: This was another issue I was seeing at the same time - entries in the languages list would jump around when filters are applied.

Because it also makes the main problem a bit easier to test, I have fixed it at the same time.

jamieparkinson (approved):
Is there work to be done here to delete old code or will that come later?

search/src/main/scala/weco/api/search/models/Aggregation.scala
```diff
@@ -59,6 +57,16 @@ object AggregationMapping {
...
   private val globalAggBuckets = root.each.buckets.each.as[Bucket]
 
+  // When we use the self aggregation pattern, buckets are returned
+  // in aggregations at multiple depths. This will return any
+  // buckets that can be found within the given Json object
+  private def bucketsFromAnywhere(json: Json): Seq[Bucket] = {
+    val allBuckets = json.findAllByKey("buckets")
+    allBuckets flatMap {
+      _.asArray.get
+    } map (k => k.as[Bucket].right.get) distinct
```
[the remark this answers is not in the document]
paul-butcher: Good point.  I think they are both 99% safe.  To break it, we would have to be passing in either a Json with a "buckets" property that does not contain a list, or that members of that list are not buckets.

I think it's guaranteed by the general context that the object being passed in conforms, but I've done nothing in here to ensure it.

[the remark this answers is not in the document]
jamieparkinson: It is my personal preference to be pedantic (quelle surprise) and only to do type-safe getting of things out of containers but up to you really

PR #687, opened 2023-08-03
Image query with vector

agnesgaroux:
## Pull request checklist

Looking at the above, it looks like it can stay as is. Maybe just change colors to singular? The API will now error if the query string is not a *single* valid hex color


search/src/main/scala/weco/api/search/models/RgbColor.scala
```diff
@@ -0,0 +1,18 @@
+package weco.api.search.models
+
+import scala.util.Try
+//import java.awt.Color
```
jamieparkinson, line 4:
```suggestion
```

jamieparkinson (approved):
🎉

search/src/main/scala/weco/api/search/models/RgbColor.scala
agnesgaroux: Gone!

PR #697, opened 2023-08-09
[WIP] start running some tests

alexwlchan:
## Pull request checklist


alexwlchan:
Closing as “who knows what I was thinking when I did this” – clearly I was deleting tests in an attempt to isolate a particular behaviour, but without any record of what it was it’s not like this can go anywhere. 🤷‍♀️

PR #718, opened 2023-09-15
Use the new rank tests

jamieparkinson:
This now runs all rank tests using the new tool that lives at https://github.com/wellcomecollection/rank


paul-butcher (approved):
So much debugged code!

PR #731, opened 2023-12-07
Some more query tweaks

jamieparkinson:
Context in Slack: https://wellcome.slack.com/archives/C3TQSF63C/p1701094384385999


paul-butcher (approved):
There's nothing wrong with having short `_name`s in the query parts, and it can help with understanding how a particular term was matched.  (not suggesting that you put them back in right now, just that it might be good to have them).

PR #736, opened 2024-01-11
Add option for HTTP healthcheck, enable for works

kenoir:
## What does this change?

This change adds the option to select a [HTTP healthcheck for a services target group](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html). The reason for this change is to avoid downtime during deploys, as at present the TCP healthcheck will only wait for nginx to be able to open a connection although the underlying scala service may not yet have started.

By default the HTTP healthcheck uses the `/management/healthcheck` endpoint (already present in the works service), and will register as healthy if that endpoint returns a 200 (see https://api.wellcomecollection.org/catalogue/v2/management/healthcheck).

> [!NOTE] 
> The intention is to roll this our for the catalogue-api works service to start with, and then add the necessary endpoints to the items and concepts service, finally removing the option for a TCP health-check. In addition further work is required to properly report the health of the service as being able to respond from the app container isn't definitive proof the app is "healthy".

## How can we test?

- [x] Manually check this healthcheck behaves as expected in the stage environment and tasks register as healthy
- [x] Perform a stage deployment while sending requests to understand if we have eliminated the deployment errors.

Changing the health-checks changes the failure modes for the API, we should test thoroughly in stage before deploying to prod, consider _and document_ the impact of extending the health check to fail in other situations (e.g. elasticsearch is unavailable).

### DIFF
--- a/terraform/modules/service/variables.tf
+++ b/terraform/modules/service/variables.tf
@@ -72,3 +72,15 @@ variable "app_cpu" {
 variable "app_memory" {
   type = number
 }
+
+variable "tcp_healthcheck" {
+  # TODO: Remove this when all services use HTTP healthcheck
+  type    = bool
+  default = true
+}
+
+variable "healthcheck_path" {
+  # Note: this is only used when tcp_healthcheck is set to false
+  type    = string
+  default = "/management/healthcheck"
+}
--- a/terraform/modules/service/variables.tf
+++ b/terraform/modules/service/variables.tf
@@ -84,3 +84,4 @@ variable "healthcheck_path" {
   type    = string
   default = "/management/healthcheck"
 }
+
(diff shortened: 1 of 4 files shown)


terraform/modules/service/variables.tf
```diff
@@ -72,3 +72,15 @@ variable "app_cpu" {
...
 }
+
+variable "tcp_healthcheck" {
+  # TODO: Remove this when all services use HTTP healthcheck
+  type    = bool
+  default = true
+}
+
+variable "healthcheck_path" {
+  # Note: this is only used when tcp_healthcheck is set to false
```
jamieparkinson, line 83: do you need both variables? ie TCP healthcheck if no path is specified, HTTP otherwise

jamieparkinson (approved):

terraform/modules/service/variables.tf
kenoir: I think if we want a default healthcheck path (which is where i'd like to end up), we need 2 variables.

PR #739, opened 2024-01-14
Adds a simple healthcheck endpoint for the items service

kenoir:
## What does this change?

This change follows https://github.com/wellcomecollection/catalogue-api/pull/736, and adds an HTTP healthcheck to the items API to ensure the scala service has started before it is registered healthy at the NLB and starts serving requests.

## How to test?

- [x] ~Manually check this healthcheck behaves as expected locally running the items API~ edit: we've improved the testing locally situation but not resolved it, I feel ok to try this in stage for now
- [ ] Deploy this change to stage to ensure the terraform applies the change as expected and tasks register as healthy
- [ ] Perform a stage deployment while sending requests to understand if we have eliminated the deployment errors.

Currently we do see _only_ items errors during deployment of the catalogue API, see from #wc-platform-alerts in slack during a deployment following updating the search :

Screenshot 2024-01-14 at 15 05 34
service healthcheck:

Changing the health-checks changes the failure modes for the API, although we have tested this principle successfully in https://github.com/wellcomecollection/catalogue-api/pull/736 so the risk is reduced.

### DIFF
--- a/items/src/main/scala/weco/api/items/ItemsApi.scala
+++ b/items/src/main/scala/weco/api/items/ItemsApi.scala
@@ -36,6 +36,14 @@ class ItemsApi(
         case id =>
           notFound(s"Work not found for identifier $id")
       }
+    },
+    pathPrefix("management") {
+      concat(
+        path("healthcheck") {
+          get {
+            complete("message" -> "ok")
+          }
+        })
     }
   )
 }
--- a/items/src/main/scala/weco/api/items/ItemsApi.scala
+++ b/items/src/main/scala/weco/api/items/ItemsApi.scala
@@ -38,12 +38,11 @@ class ItemsApi(
       }
     },
     pathPrefix("management") {
-      concat(
-        path("healthcheck") {
-          get {
-            complete("message" -> "ok")
-          }
-        })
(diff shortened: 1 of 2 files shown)


agnesgaroux:
Not sure how this works. Does it check /management/healthcheck before hitting /works every time? Then if the healthcheck fails (for whatever reason, could be something other than the instance being currently in deployment) the LB what? tries another instance?

kenoir:
> Not sure how this works. Does it check /management/healthcheck before hitting /works every time? Then if the healthcheck fails (for whatever reason, could be something other than the instance being currently in deployment) the LB what? tries another instance?

This change only provides a new endpoint at `/management/healthcheck` that serves the following json and doesn't do anything else:

```json
{
  "message": "ok"
}
```

The load balancer uses this endpoint to determine if the instance is healthy, and if it is allow it to serve requests. At present the load balancer uses a TCP health check that only relies on the nginx sidecar that proxies requests to the app to be available. Nginx comes up very quick while the slowpoke scala app is still yawning and blinking itself awake. This change makes sure the scala app is up by forcing it to serve requests before the load balancer determines it to b healthy.

PR #747, opened 2024-01-25
Adds a simple healthcheck endpoint for the requests service

kenoir:
## What does this change?

This change follows https://github.com/wellcomecollection/catalogue-api/pull/739, and adds an HTTP healthcheck to the requests API to ensure the Scala service has started before it is registered healthy at the NLB and starts serving requests.

A further [change will need to be made in the wellcomecollection/identity](https://github.com/wellcomecollection/identity/blob/main/infra/modules/ecs_service/target_group.tf) respository to have the identity infrastructure pay attention to the new endpoint. 

The required change will be similar to the one in [this repository to add a HTTP healthcheck](https://github.com/wellcomecollection/catalogue-api/pull/736).

See https://github.com/wellcomecollection/identity/pull/398

## How to test?

- [ ] Perform a stage deployment while sending requests to understand if we have eliminated any deployment errors.

### DIFF
--- a/concepts/src/controllers/healthcheck.ts
+++ b/concepts/src/controllers/healthcheck.ts
@@ -1,0 +1,16 @@
+import { RequestHandler } from "express";
+import asyncHandler from "express-async-handler";
+import { Config } from "../../config";
+
+type PathParams = { id: string };
+
+const healthcheckController = (config: Config): RequestHandler<PathParams> => {
+  return asyncHandler(async (req, res) => {
+    res.status(200).json({
+      status: "ok",
+      config,
+    });
+  });
+};
+
+export default healthcheckController;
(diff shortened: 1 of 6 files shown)


rcantin-w (approved):

concepts/src/controllers/healthcheck.ts
```diff
@@ -0,0 +1,16 @@
...
+import asyncHandler from "express-async-handler";
+import { Config } from "../../config";
+
+type PathParams = { id: string };
+
+const healthcheckController = (config: Config): RequestHandler<PathParams> => {
+  return asyncHandler(async (req, res) => {
+    res.status(200).json({
+      status: "ok",
+      config,
```
paul-butcher, line 11: When I saw this, I was slightly concerned that it might be a vector for leaking secrets.

It isn't, but is there a risk that someone might unwittingly expose something in the future by adding it to the config?

paul-butcher (approved):

PR #749, opened 2024-01-25
Adds a simple healthcheck endpoint for the concepts service

kenoir:
## What does this change?

This change follows https://github.com/wellcomecollection/catalogue-api/pull/747, and adds an HTTP healthcheck to the requests API to ensure the Scala service has started before it is registered healthy at the NLB and starts serving requests.

## How to test?

- [ ] Perform a stage deployment while sending requests to understand if we have eliminated any deployment errors.

### DIFF
--- a/terraform/modules/service/target_group.tf
+++ b/terraform/modules/service/target_group.tf
@@ -15,22 +15,10 @@ resource "aws_lb_target_group" "tcp" {
   # updated service.  Reducing this parameter to 90s makes deployments faster.
   deregistration_delay = 90
 
-  dynamic "health_check" {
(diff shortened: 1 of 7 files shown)


terraform/modules/service/target_group.tf
```diff
@@ -15,22 +15,10 @@ resource "aws_lb_target_group" "tcp" {
   # updated service.  Reducing this parameter to 90s makes deployments faster.
   deregistration_delay = 90
 
-  dynamic "health_check" {
```
kenoir, line 18: Addresses https://github.com/wellcomecollection/catalogue-api/pull/736#discussion_r1450363164

agnesgaroux (approved):

PR #754, opened 2024-02-19
new pipeline 02-09

paul-butcher:
This updates the pipeline version to associate the API with a pipeline that has been populated using the [two-stage not-Miro-then-Miro process](https://github.com/wellcomecollection/catalogue-pipeline/pull/2553).  

This ensures that when a digmiro image hits the merger, it is always merged correctly.

Previously, digmiro images were being erroneously created as standalone images when the micro record hit the merger before the corresponding Sierra record.


paul-butcher:
superseded

PR #756, opened 2024-03-01
return items with available dates if applicable

agnesgaroux:
This returns a hardcoded list of `AvailabilitySlots` with items that satisfying the [conditions](https://github.com/wellcomecollection/catalogue-api/blob/dc94db57e042d6b92b472c1d9ffee2fb81add3da/common/stacks/src/main/scala/weco/api/stacks/models/DisplayItemOps.scala#L71)

NOTE: the last case in the list of `itemStates` should have no available dates:
```
missingItemResponse(workWithAvailableItemNumber),
workWithAvailableItem,
availableOnlineAccessCondition,
availableDates // this should be None, possible bug in the Sierra updater
```
It looks like there's a bug with how the missing items are handled: we keep them in the item list without updating their accessConditions. 

I was trying to reinvent the wheel until I realised there's already a `DisplayAccessConditionOps` class that give you a boolean "isRequestable" from the item's `DisplayAccessCondition`. Might be worth double checking the logic is sound in there

## Pull request checklist

### DIFF
--- a/items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
+++ b/items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
@@ -15,7 +15,6 @@ import weco.sierra.models.fields.SierraItemDataEntries
 import weco.sierra.models.identifiers.SierraItemNumber
 
 import scala.concurrent.{ExecutionContext, Future}
-import java.time.ZonedDateTime
 
 /** Updates the AccessCondition of sierra items
   *
@@ -61,11 +60,9 @@ class SierraItemUpdater(sierraSource: SierraSource)(
     for {
       itemEither <- sierraSource.lookupItemEntries(existingItems.keys.toSeq)
 
-      maybeAccessConditions: Map[
-        SierraItemNumber,
-        Option[
-          DisplayAccessCondition
-        ]] = itemEither match {
+      maybeAccessConditions: Map[SierraItemNumber, Option[
+        DisplayAccessCondition
+      ]] = itemEither match {
         case Right(SierraItemDataEntries(_, _, entries)) =>
           entries
             .map(item => {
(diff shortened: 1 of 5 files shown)


items/src/test/scala/weco/api/items/services/ItemUpdateServiceTest.scala
```diff
@@ -150,8 +151,17 @@ class ItemUpdateServiceTest
     status = CatalogueAccessStatus.Open
   )
 
-  val notRequestableAccessCondition = DisplayAccessCondition(
-    method = CatalogueAccessMethod.NotRequestable
+  val availableDates = Some(
+    List(
+      AvailabilitySlot(
+        ZonedDateTime.parse(
+          "2024-02-29T13:32:44.943107Z[Europe/London]"
```
jamieparkinson, line 158: I think - but I'm definitely open to having my mind changed - the API should return _either_ a UTC ISO8601 timestamp, or an ISO8601 timestamp with an hour offset. The timezone name in square brackets thing is non-standard and so I feel it's best to avoid exposing it... Maybe start a thread in Slack to share the question with the team?

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -121,6 +121,31 @@ class SierraItemUpdater(sierraSource: SierraSource)(
...
+                        .parse("2024-02-29T13:32:44.943107Z[Europe/London]")
+                        .plusWeeks(2)
+                    )
+                  )
+                )
+              )
+            case false => item
+          }
+      }
+    } yield updatedItemsWithAvailableDates.toSeq
```
jamieparkinson, line 149: In anticipation of the next stage of work, I would suggest splitting out the addition of available dates into another class or something - so that no changes would be required here when we use "real" dates

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -121,6 +121,31 @@ class SierraItemUpdater(sierraSource: SierraSource)(
             case None            => item
           }
       }
-    } yield updatedItems.toSeq
+
+      updatedItemsWithAvailableDates = updatedItems.map {
+        case (item) =>
+          val displayAccessConditionOps = new DisplayAccessConditionOps(
+            item.locations.head.accessConditions.head
```
jamieparkinson, line 128: Because this is an implicit class, so long as it's imported you don't need to construct it: `isRequestable` will just appear on any instance of `DisplayAccessCondition`! https://docs.scala-lang.org/overviews/core/implicit-classes.html#usage

items/src/test/scala/weco/api/items/services/ItemUpdateServiceTest.scala
agnesgaroux: That was a bit of an oversight, I agree with UTC ISO8601 timestamp 👍 
That weird format is java ZonedDateTime

jamieparkinson (approved):

Issue #759, opened 2024-03-07
Guard against "Head of empty list" errors in Sierraitemupdater


paul-butcher:
We keep getting alerted to 5xx errors occurring when an item has no accessConditions. 
https://wellcome.slack.com/archives/CQ720BG02/p1709821702412539

https://wellcome.slack.com/archives/CQ720BG02/p1709817801997699

pollecuttn:
I can't see status or access - is that my internet or a problem for us?
![image](https://github.com/wellcomecollection/catalogue-api/assets/1446072/87cd71d1-fdeb-4c07-97ad-105e92ef249d)

pollecuttn:
But I can see them on the OPAC:
![image](https://github.com/wellcomecollection/catalogue-api/assets/1446072/341f64e7-9bd1-4174-be17-b35934793fc0)

paul-butcher:
> I can't see status or access - is that my internet or a problem for us?

Yes, I think that is our problem. I think this will fix that, but there may be something sending us dodgy data elsewhere.

PR #760, opened 2024-03-07
guard against getting stuff out of an empty list

paul-butcher:
Fixes https://github.com/wellcomecollection/catalogue-api/issues/759

This is a fairly quick fix, to stop these errors happening, but I would like us to revisit this section of the codebase again soon. 

Although the fix was pretty simple to write, the only existing tests that exercise this function are a bit unwieldy and end-to-endy.  `updateItems` is not very testable and requires a load of context manager setup and boilerplate, when it would be preferable to break it up into a function that takes some inputs and produces some outputs.  Then it would be much simpler to define a more comprehensive suite of tests that exercise all the possibilities.


agnesgaroux (approved):
whoops :( 
So we're sending back items without their location then? I've seen it on wc.org i think, it looks like it's loading forever

kenoir:
> whoops :( So we're sending back items without their location then? I've seen it on wc.org i think, it looks like it's loading forever

I think i've seen this too.

Issue #767, opened 2024-03-21
Get nextOpeningDates from content-api to return availableDates from itemsAPI


agnesgaroux:
Following on from https://github.com/wellcomecollection/catalogue-api/issues/741

jcateswellcome:
part of #784

PR #771, opened 2024-04-25
Get next opening dates and add to response

agnesgaroux:
## Pull request checklist
https://github.com/wellcomecollection/catalogue-api/issues/767 

[swagger]: https://github.com/wellcomecollection/developers.wellcomecollection.org/blob/main/reference/catalogue.yaml Maybe? `'/works/{id}/items` is not documented. Deliberately?

### DIFF
--- a/items/src/main/scala/weco/api/items/Main.scala
+++ b/items/src/main/scala/weco/api/items/Main.scala
@@ -22,11 +22,8 @@ import weco.sierra.http.SierraSource
 import weco.sierra.typesafe.SierraOauthHttpClientBuilder
 
 import scala.concurrent.ExecutionContext
-
-trait Clock {
-  def now(): ZonedDateTime
-}
-class LondonClock extends Clock {
+//}
+class LondonClock {
   def now(): ZonedDateTime =
     ZonedDateTime.now.withZoneSameLocal(ZoneId.of("Europe/London"))
   def getHour: Int = now().getHour
--- a/items/src/main/scala/weco/api/items/Main.scala
+++ b/items/src/main/scala/weco/api/items/Main.scala
@@ -7,6 +7,7 @@ import weco.Tracing
 import weco.api.items.services.{
   ItemUpdateService,
   SierraItemUpdater,
+  VenueOpeningTimesLookup,
   WorkLookup
 }
 import weco.http.typesafe.HTTPServerBuilder
@@ -37,21 +38,26 @@ object Main extends WellcomeTypesafeApp {
     val client = SierraOauthHttpClientBuilder.build(config)
     val sierraSource = new SierraSource(client)
 
+    val contentHttpClient = new AkkaHttpClient() with HttpGet {
+      override val baseUri: Uri = config.getString("content.api.publicRoot")
+    }
+    val venueOpeningTimeLookup = new VenueOpeningTimesLookup(contentHttpClient)
+
     // To add an item updater for a new service:
     // implement ItemUpdater and add it to the list here
     val itemUpdaters = List(
-      new SierraItemUpdater(sierraSource)
+      new SierraItemUpdater(sierraSource, venueOpeningTimeLookup)
     )
 
     val itemUpdateService = new ItemUpdateService(itemUpdaters)
 
-    val httpClient = new AkkaHttpClient() with HttpGet {
+    val catalogueHttpClient = new AkkaHttpClient() with HttpGet {
       override val baseUri: Uri = config.getString("catalogue.api.publicRoot")
     }
 
     val router = new ItemsApi(
       itemUpdateService = itemUpdateService,
-      workLookup = new WorkLookup(httpClient)
+      workLookup = new WorkLookup(catalogueHttpClient)
     )
 
     val appName = "ItemsApi"
--- a/items/src/main/scala/weco/api/items/Main.scala
+++ b/items/src/main/scala/weco/api/items/Main.scala
@@ -43,7 +43,7 @@ object Main extends WellcomeTypesafeApp {
       override val baseUri: Uri = config.getString("content.api.publicRoot")
     }
     val venueOpeningTimeLookup = new VenueOpeningTimesLookup(contentHttpClient)
-    val clock = Clock.system(ZoneId.of("Europe/London"))
+    val venueClock = Clock.system(ZoneId.of("Europe/London"))
 
     // To add an item updater for a new service:
     // implement ItemUpdater and add it to the list here
@@ -51,7 +51,7 @@ object Main extends WellcomeTypesafeApp {
       new SierraItemUpdater(
         sierraSource,
         venueOpeningTimeLookup,
-        clock
+        venueClock
       )
     )
 
--- a/items/src/main/scala/weco/api/items/Main.scala
+++ b/items/src/main/scala/weco/api/items/Main.scala
@@ -19,8 +19,8 @@ import weco.http.client.{AkkaHttpClient, HttpGet}
 import weco.http.monitoring.HttpMetrics
 import weco.sierra.http.SierraSource
 import weco.sierra.typesafe.SierraOauthHttpClientBuilder
-import weco.api.items.services.LondonClock
 
+import java.time.{Clock, ZoneId}
 import scala.concurrent.ExecutionContext
 
 object Main extends WellcomeTypesafeApp {
@@ -43,6 +43,7 @@ object Main extends WellcomeTypesafeApp {
       override val baseUri: Uri = config.getString("content.api.publicRoot")
     }
     val venueOpeningTimeLookup = new VenueOpeningTimesLookup(contentHttpClient)
+    val clock = Clock.system(ZoneId.of("Europe/London"))
 
     // To add an item updater for a new service:
     // implement ItemUpdater and add it to the list here
@@ -50,7 +51,7 @@ object Main extends WellcomeTypesafeApp {
       new SierraItemUpdater(
         sierraSource,
         venueOpeningTimeLookup,
-        new LondonClock()
+        clock
       )
     )
 
--- a/items/src/main/scala/weco/api/items/Main.scala
+++ b/items/src/main/scala/weco/api/items/Main.scala
@@ -2,6 +2,7 @@ package weco.api.items
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.Uri
+import java.time.{ZoneId, ZonedDateTime}
 import com.typesafe.config.Config
 import weco.Tracing
 import weco.api.items.services.{
@@ -21,6 +22,15 @@ import weco.sierra.http.SierraSource
 import weco.sierra.typesafe.SierraOauthHttpClientBuilder
 
 import scala.concurrent.ExecutionContext
+
+trait Clock {
+  def now(): ZonedDateTime
+}
+class LondonClock extends Clock {
+  def now(): ZonedDateTime =
+    ZonedDateTime.now.withZoneSameLocal(ZoneId.of("Europe/London"))
+  def getHour: Int = now().getHour
+}
 
 object Main extends WellcomeTypesafeApp {
 
@@ -46,7 +56,11 @@ object Main extends WellcomeTypesafeApp {
     // To add an item updater for a new service:
     // implement ItemUpdater and add it to the list here
     val itemUpdaters = List(
-      new SierraItemUpdater(sierraSource, venueOpeningTimeLookup)
+      new SierraItemUpdater(
+        sierraSource,
+        venueOpeningTimeLookup,
+        new LondonClock()
+      )
     )
 
     val itemUpdateService = new ItemUpdateService(itemUpdaters)
--- a/items/src/main/scala/weco/api/items/Main.scala
+++ b/items/src/main/scala/weco/api/items/Main.scala
@@ -2,7 +2,6 @@ package weco.api.items
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.Uri
-import java.time.{ZoneId, ZonedDateTime}
 import com.typesafe.config.Config
 import weco.Tracing
 import weco.api.items.services.{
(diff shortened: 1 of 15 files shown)


items/src/main/scala/weco/api/items/Main.scala
```diff
@@ -20,6 +22,12 @@ import weco.sierra.http.SierraSource
 import weco.sierra.typesafe.SierraOauthHttpClientBuilder
 
 import scala.concurrent.ExecutionContext
+//}
```
jamieparkinson, line 25:
```suggestion
```

items/src/test/scala/weco/api/items/services/ItemUpdateServiceTest.scala
```diff
@@ -270,94 +301,167 @@ class ItemUpdateServiceTest
...
       (
         "Sierra Responses",
+        "Content api Venue Responses",
         "Catalogue Work",
         "AccessCondition",
         "AvailableDates"
       ),
       (
         onHoldItemResponses(workWithAvailableItemNumber),
+        Seq(),
```
jamieparkinson, line 311: Scala trivia: as well as `Seq()` or `List()` or `Seq.empty` or `List.empty` and so on, you can also just use `Nil` for any empty iterable

items/src/main/scala/weco/api/items/Main.scala
```diff
@@ -20,6 +22,12 @@ import weco.sierra.http.SierraSource
...
+class LondonClock {
+  def now(): ZonedDateTime =
+    ZonedDateTime.now.withZoneSameLocal(ZoneId.of("Europe/London"))
+  def getHour: Int = now().getHour
+}
```
jamieparkinson, line 30: Couple of things about this class; firstly I would move it to its own file (probably in `services`) as you're also using it elsewhere.

When we define it as a `LondonClock` here, name it in a parameter as `clock`, and then write`clock.now()` and `clock.getHour` we kind of lose the context of the timezone: in `SierraItemUpdater` it's not clear what time the clock is giving you. The important thing is that this clock gives you the time at Wellcome Collection.

Maybe we could do something like this:
```scala
class VenueClock {
  def timeAtVenue: ZonedDateTime = ??? // London time
  def hourAtVenue: Int = ??? // as above
}
```

then when we use it, it's always clear where the time is coming from: `clock.timeAtVenue`

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -92,17 +97,53 @@ class SierraItemUpdater(sierraSource: SierraSource)(
...
-      )
+  private def setAvailableDates(
+    item: DisplayItem
+  ): Future[DisplayItem] = {
+    // there is only ever one location per physicalItem and one accessCondition per location,
+    // but this may sometimes be empty if it could not be fetched.
+    val accessCondition = item.locations.headOption
+      .flatMap(_.accessConditions.headOption)
+
+    if ((accessCondition.isDefined && accessCondition.head.isRequestable)) {
```
jamieparkinson, line 108:
```suggestion
    if (accessCondition.exists(_.isRequestable)) {
```

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -92,17 +97,53 @@ class SierraItemUpdater(sierraSource: SierraSource)(
...
+  ): Future[List[AvailabilitySlot]] = {
+    val venueFromDisplayAccessCondition = Map(
+      "online-request" -> "library"
+      // other venue to be added as DisplayAccessMethod id -> content-api venue title
     )
+    val hourNow = clock.getHour
+    val leadTime = accessCondition.method.id match {
+      case "online-request" if hourNow < 10  => 1
+      case "online-request" if hourNow >= 10 => 2
+    }
```
jamieparkinson, line 127: This is a tricky one - the `leadTime` here isn't a time, it's a number of availability slots that we're going to `drop`. We should probably rename it something like `leadTimeDays`, but how does the items API know that an availability slot is 1 day? Should it be checking that dropping this number of slots corresponds to a lead time of this number of days? See comment below for a suggestion!

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -92,17 +97,53 @@ class SierraItemUpdater(sierraSource: SierraSource)(
...
+      .byVenueName(
+        venueFromDisplayAccessCondition.get(accessCondition.method.id)
+      )
+      .map {
+        case Right(venue) =>
+          venue.openingTimes
+            .map(
+              openClose => AvailabilitySlot(openClose.open, openClose.close)
+            )
+            .drop(leadTime)
```
jamieparkinson, line 139:
```suggestion
        case Right(venue) =>
          venue.openingTimes
            .map(
              openClose => AvailabilitySlot(openClose.open, openClose.close)
            )
            .dropWhile(slot => daysAwayFromNow(slot) < leadTimeDays)
```

This suggestion assumes the existence of a new method that might look a bit like this:
```scala
def daysAwayFromNow(slot: AvailabilitySlot): Int = {
  val closeTime = ZonedDateTime.parse(slot.close)
  ChronoUnit.between(clock.timeAtVenue, closeTime, ChronoUnit.DAYS)
}
```

items/src/main/scala/weco/api/items/services/VenueOpeningTimesLookup.scala
```diff
@@ -0,0 +1,73 @@
...
+  implicit as: ActorSystem,
+  ec: ExecutionContext
+) extends Logging {
+
+  implicit val um: Unmarshaller[HttpEntity, ContentApiVenueResponse] =
+    CirceMarshalling.fromDecoder[ContentApiVenueResponse]
+
+  /** Returns the Work that corresponds to this canonical ID.
+    *
+    */
```
jamieparkinson, line 32: Think this comment needs updating!

jamieparkinson (commented):
Very nice, just have a few suggestions

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -92,18 +98,59 @@ class SierraItemUpdater(sierraSource: SierraSource)(
...
     )
 
+    val hourAtVenue = clock.timeInLondon().getHour
+    val leadTimeInDays = accessCondition.method.id match {
+      case "online-request" if hourAtVenue < 10  => 1
+      case "online-request" if hourAtVenue >= 10 => 2
+    }
+    def daysAwayFromNow(slot: AvailabilitySlot): Int = {
+      val closeTime = ZonedDateTime.parse(slot.to)
+      ChronoUnit.DAYS.between(clock.timeInLondon(), closeTime).toInt
```
jamieparkinson, line 132: Nit: if we're defining this method inline, then it makes sense to close over the time (ie bind the variable from the outer scope rather than asking `clock` again)

```suggestion
    val timeAtVenue = clock.timeInLondon()
    val leadTimeInDays = accessCondition.method.id match {
      case "online-request" if timeAtVenue.getHour < 10  => 1
      case "online-request" if timeAtVenue.getHour >= 10 => 2
    }
    def daysAwayFromNow(slot: AvailabilitySlot): Int = {
      val closeTime = ZonedDateTime.parse(slot.to)
      ChronoUnit.DAYS.between(timeAtVenue, closeTime).toInt
```

Also (more importantly actually) - I think I might've been wrong about using `closeTime` here; should it be the opening time instead?

jamieparkinson (approved):
Nice, approving it but have an outstanding comment that my suggestion might not have used the right "edge" of the availability slot

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -22,7 +24,11 @@ import scala.concurrent.{ExecutionContext, Future}
   *  can be placed on an item.
   *
   */
-class SierraItemUpdater(sierraSource: SierraSource)(
+class SierraItemUpdater(
+  sierraSource: SierraSource,
+  venueOpeningTimesLookup: VenueOpeningTimesLookup,
+  clock: LondonClock
```
kenoir, line 30: I'd be tempted to make use of a [java `Clock`](https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html) here, which in tests can be provided [as a fixed time](https://mincong.io/2020/05/24/java-clock/).

items/src/test/scala/weco/api/items/fixtures/ItemsApiFixture.scala
```diff
@@ -30,12 +33,28 @@ trait ItemsApiFixture extends SierraSourceFixture {
 
   def withItemsApi[R](
     catalogueResponses: Seq[(HttpRequest, HttpResponse)] = Seq(),
-    sierraResponses: Seq[(HttpRequest, HttpResponse)] = Seq()
+    sierraResponses: Seq[(HttpRequest, HttpResponse)] = Seq(),
+    contentApiVenueResponses: Seq[(HttpRequest, HttpResponse)] = Seq(),
+    time: Int
```
kenoir, line 38: It might be clearer in tests what's going on if this took a `Clock` and there was another `withClock` function to wrap creating that clock (i'd also provide a default here for tests that do not care about the time).

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
agnesgaroux: I think close v open doesn't really matter, `ChronoUnit.DAYS.between` only looks at the day which is the same for open and close (these places are not open overnight)

jamieparkinson: You and I know that - but how does the application? There's nothing in the model that tells us a slot only covers a day (the same reason that I suggested making this change to `dropWhile`). If we write this in a way that doesn't make assumptions about the data, then it means that someone reading it in future doesn't have to know that context, and also makes it robust to any potential changes in the data.

I think it needs to use the opening time, as that means the lead time is sufficient that the item will be available if the user turns up right at the start of the slot - does that sound right?

items/src/main/scala/weco/api/items/Main.scala
agnesgaroux: Replaced by java Clock

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
agnesgaroux: Now using dropWhile, added a comment to explain why we're dropping slots too

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
agnesgaroux: Done and done :)

items/src/test/scala/weco/api/items/services/ItemUpdateServiceTest.scala
```diff
@@ -38,30 +38,32 @@ class ItemUpdateServiceTest
...
             sierraSource,
             new VenueOpeningTimesLookup(contentApiClient),
-            londonClock
+            clock
           )
         )
       }
     }
 
+  def withClock(time: Int = 11): Clock = {
```
kenoir, line 60: I would implement this using the `TestWith` pattern like this: https://github.com/wellcomecollection/catalogue-api/blob/5cd4f66dcc362aca2fb60baa13349623d68d2d46/items/src/test/scala/weco/api/items/services/ItemUpdateServiceTest.scala#L36

You could have it take a parseable date time string:

```scala
import import weco.fixtures.TestWith

def withClock[R](
  dateTime: String = "2018-04-29T10:15:30.00Z"
)(testWith: TestWith[Clock, R]): R = {
  val instant = Instant.parse(dateTime)
  val zoneId = ZoneId.of("Europe/London")
  testWith(Clock.fixed(instant, zoneId))
}
```

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -22,7 +24,11 @@ import scala.concurrent.{ExecutionContext, Future}
   *  can be placed on an item.
   *
   */
-class SierraItemUpdater(sierraSource: SierraSource)(
+class SierraItemUpdater(
+  sierraSource: SierraSource,
+  venueOpeningTimesLookup: VenueOpeningTimesLookup,
+  clock: Clock
```
jamieparkinson, line 30:
```suggestion
  venueClock: Clock
```
Agreed on using a `Clock` but still important not to lose context of what it's doing at the point of use

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -92,18 +98,62 @@ class SierraItemUpdater(sierraSource: SierraSource)(
...
+      // other venue to be added as DisplayAccessMethod id -> content-api venue title
     )
 
+    val timeAtVenue = LocalDateTime.now(clock)
+    val leadTimeInDays = accessCondition.method.id match {
+      case "online-request" if timeAtVenue.getHour < 10  => 1
+      case "online-request" if timeAtVenue.getHour >= 10 => 2
+    }
+    def daysAwayFromNow(slot: AvailabilitySlot): Int = {
+      val openingTime = ZonedDateTime.parse(slot.to)
```
jamieparkinson, line 131: This is the closing time, isn't it?

agnesgaroux: whoops

items/src/test/scala/weco/api/items/fixtures/ItemsApiFixture.scala
```diff
@@ -28,28 +30,48 @@ trait ItemsApiFixture extends SierraSourceFixture {
...
-      withSierraSource(sierraResponses) { sierraSource =>
-        val itemsUpdaters = List(
-          new SierraItemUpdater(sierraSource)
-        )
-
-        val catalogueApiClient = new MemoryHttpClient(catalogueResponses)
-        with HttpGet {
-          override val baseUri: Uri = Uri("http://catalogue:9001")
-        }
+      withClock() { clock =>
```
agnesgaroux, line 47: We don't really care about the availableDates at that level so using the default in the fixture, instead of wrapping all the tests with `withClock`

kenoir (approved):
👌

PR #774, opened 2024-04-30
Add the ability to run the search project locally

kenoir:
## What does this change?

Allow the search project to run locally, in order to better test changes by querying a configured real Elasticsearch index.

## How to test

Attempt to run this project locally, does it work?

Developers have much better understanding of how their changes will work when deployed to production.

We have changed the`application.conf` so unless parameters are overridden by env vars when deployed unexpected behaviour may occur. Looking at the terraform config for this service it appears the params for which new defaults are provided are overwritten.

### DIFF
--- a/common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
+++ b/common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
@@ -25,9 +25,10 @@ object PipelineElasticClientBuilder {
 
     val secretsManagerClientBuilder = SecretsManagerClient.builder()
 
-    implicit val secretsClient: SecretsManagerClient = if(isDev) {
+    implicit val secretsClient: SecretsManagerClient = if (isDev) {
       secretsManagerClientBuilder
-        .credentialsProvider(ProfileCredentialsProvider.create("catalogue-developer"))
+        .credentialsProvider(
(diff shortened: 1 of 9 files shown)


project/Dependencies.scala
```diff
@@ -128,6 +128,10 @@ object ExternalDependencies {
     "software.amazon.awssdk" % "secretsmanager" % versions.aws2
   )
 
+  val stsDependencies = Seq(
```
kenoir, line 131: This is required so the role assumption can work.

common/search/src/main/scala/weco/api/search/config/builders/PipelineElasticClientBuilder.scala
```diff
@@ -18,14 +19,27 @@ object PipelineElasticClientBuilder {
 
   def apply(
     serviceName: String,
-    pipelineDate: String = PipelineClusterElasticConfig.pipelineDate
+    pipelineDate: String = PipelineClusterElasticConfig.pipelineDate,
+    isDev: Boolean = false
```
jamieparkinson, line 23: Perhaps worth using an `environment` string to pattern match on instead? Generally I am opposed to boolean flags...

kenoir: Agreed.

jamieparkinson (approved):

PR #775, opened 2024-05-02
Add the ability to run the items project locally

kenoir:
## What does this change?

Allow the items project to run locally, in order to better test changes by querying a configured real Sierra.

Follows: https://github.com/wellcomecollection/catalogue-api/pull/774

## How to test

Attempt to run this project locally, does it work?

Developers have much better understanding of how their changes will work when deployed to production.

We have changed the `application.conf` so unless parameters are overridden by env vars when deployed unexpected behaviour may occur. Looking at the terraform config for this service it appears the params for which new defaults are provided are overwritten.

### DIFF
--- a/items/src/main/resources/application.conf
+++ b/items/src/main/resources/application.conf
@@ -5,7 +5,7 @@ apm.service.name=${?apm_service_name}
 apm.environment=${?apm_environment}
 
 api.public-root="http://localhost"
-api.public-root=${?api_public_root}
+api.public-root=${?catalogue_api_public_root}
 
 catalogue.api.publicRoot="https://api.wellcomecollection.org/catalogue/v2"
 catalogue.api.publicRoot=${?catalogue_api_public_root}
--- a/items/src/main/resources/application.conf
+++ b/items/src/main/resources/application.conf
@@ -4,14 +4,21 @@ apm.secret=${?apm_secret}
 apm.service.name=${?apm_service_name}
 apm.environment=${?apm_environment}
 
+api.public-root="http://localhost"
+api.public-root=${?api_public_root}
+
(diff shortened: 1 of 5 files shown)


items/src/main/resources/application.conf
```diff
@@ -4,15 +4,21 @@ apm.secret=${?apm_secret}
 apm.service.name=${?apm_service_name}
 apm.environment=${?apm_environment}
 
+api.public-root="http://localhost"
+api.public-root=${?api_public_root}
```
kenoir, line 8: This needs a terraform update to set this in an ENV variable.

items/src/main/resources/application.conf
```diff
@@ -4,15 +4,21 @@ apm.secret=${?apm_secret}
...
+
+catalogue.api.publicRoot="https://api.wellcomecollection.org/catalogue/v2"
 catalogue.api.publicRoot=${?catalogue_api_public_root}
+
+content.api.publicRoot="https://api.wellcomecollection.org/content/v0"
 content.api.publicRoot=${?content_api_public_root}
 
 http.host="0.0.0.0"
+http.port=8080
```
paul-butcher, line 17: Do we actually use port and host separately?   Below, we have the default localhost url defined literally, rather than in terms of host and port.

paul-butcher (approved):

items/src/main/resources/application.conf
kenoir: It's the other way round, we only use them separately AFAICT. In `externalBaseURL` the port gets trimmed off before it's used too so that is redundant.

kenoir: Agreed, the env vars & the way we construct the config here could probably be reconciled to something more sensible.

PR #778, opened 2024-05-07
fix lead time bug + add test case

agnesgaroux:
## Pull request checklist

*   Does this patch need a change to the documentation? No 

NOTE: uncovered an edge case for requests made on a closed day, after 10am. 
The logic previously in place dictated that we should drop the 1st and 2nd nextOpeningDays because the req is made after 10am. 
However in reality the staff will actually **get** the request before 10am on the 1st nextOpeningDay, so we should only drop the 1st nextOpeningDays. 
This is the current behaviour in weco.org


kenoir (approved):
👌

PR #791, opened 2024-06-07
Use a different record for the smoke test

paul-butcher:
The smoke tests have started failing because of a change in the underlying data.  Nature (awa6c6gm, redirecting to vruhajjv) does not have `items`, which means that the test on lines 42-48 have now started failing because "/works/awa6c6gm/items" returns a 500 error.

Leaving aside that it should probably return a client error rather than a server error, it is still necessary to change this test to use a record that returns 200 when `items` are requested.


kenoir (approved):
🩹

Issue #787, opened 2024-05-31
[offsiteReq] Update SierraItemUpdater to add availabilitySlots for DeepStore items


agnesgaroux:
Pull in the item's location code from Sierra when we get its access conditions [here](https://github.com/wellcomecollection/catalogue-api/blob/c4a062b8d5f75d839f658a61afc5846ea700337d/items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala#L64) then use it in `getAvailableDates` to figure out whether it's DeepStore (harop) or onsite (any other location code)

If it's DeepStore, fetch the venue from content-api and apply correct `LeadTimeInDays`

❓I don't think the location code should be part of the `DisplayAccessCondition` since it's not something we want to display on the work page. We could include it and let wc.org ignore it, or refactor `updateItems` so that it handles items one by one which I believe would keep things are bit cleaner 

-> there will be some slight complication around the available dates **after** the first
- first available date is dependent on DeppStore working days. Alexandra confirmed it's [any day other than Saturdays, Sundays and public holidays](https://wellcome.slack.com/archives/C05JE6PE5PG/p1716999415777119?thread_ts=1716998691.301479&cid=C05JE6PE5PG)
- the following days are dependent on the library opening days: once the item is onsite it can be viewed in the library any day it is open
- Therefore we'll need to fetch both DeeptStore and Library venues for DeepStore items. The 1st `AvailabilitySlot` will be the day the item is delivered, then the following ones will be library opening days

agnesgaroux:
Rules for DeepStore leadTimeInDays

The logic as I understand it is:
- 10 DeepStore working days to get the item back on site
- After that the item can be viewed any day that the library is open

If I make a request on Monday 3rd June:
- the item will be delivered on site on Friday 14th June
- it can be viewed Saturday 15th June, then Monday 17th to Saturday 22nd and Monday 24th to Friday 28th (the current choice of 12 days to pick from)
- If there were x public holidays between Monday 3rd June and Friday 14th June, the delivery date would be postponed by as many days

That's assuming, TBC, that:
- the day of the request (Monday 3rd June) is day 1 of the 10 days DeepStore lead time, not day 0
- the item cannot be viewed on the day it is delivered, but the next day the library is open
- we give the user a 12 opening day window to come on site to view the item once it's been delivered. If they want to see it at a later date, they ned to make the request at a later date

agnesgaroux:
Email sent to LE&E to confirm the above points

[...]

PR #794, opened 2024-07-11
Integrate deepstore into items api sierra item updater

agnesgaroux:
## Pull request checklist

Interesting changes are in items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala

- Needs tests
- Type mismatch in `updateItem`: some values are `Option[value]` but the `value` is expected (see failed build in BK). Trying to find an elegant way around that


items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -57,142 +103,123 @@ class SierraItemUpdater(
...
-        case "online-request" if (timeAtVenue.getHour < 10 || !isWorkingDay) =>
-          1
-        case "online-request" if timeAtVenue.getHour >= 10 => 2
-      }
-    }
+  private def getVenuesOpeningTimes(
+    locationName: String
+  ): Future[Map[String, List[AvailabilitySlot]]] =
+    for {
+      venues <- venuesOpeningTimesLookup
```
kenoir, line 149: I'd avoid the `for` comprehension here as you have only one future to unwrap (and it'll unnest things a bit here).

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -22,22 +21,69 @@ import scala.concurrent.{ExecutionContext, Future}
...
+  val identifierType: DisplayIdentifierType =
+    DisplayIdentifierType.SierraSystemNumber
 
-  val identifierType = DisplayIdentifierType.SierraSystemNumber
+  private def lookupItems(
+    staleItemIds: Seq[SierraItemNumber]
+  ): Future[Map[SierraItemNumber, SierraItemData]] = {
+    val itemsEither = sierraSource.lookupItemEntries(staleItemIds)
+
+    itemsEither.map(itemEither => {
```
kenoir, line 46: Nitypicky:

```suggestion
  ): Future[Map[SierraItemNumber, SierraItemData]] = 
    sierraSource.lookupItemEntries(staleItemIds).map(itemEither => {
```

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -57,142 +103,123 @@ class SierraItemUpdater(
...
+                    AvailabilitySlot(openingTime.open, openingTime.close)
             )
-            // the list of openingTimes, as returned from VenueOpeningTimesLookup, starts at "closest open day"
-            // (including today)
-            // however, it takes ${leadTimeInDays} days for the item to be fetched from stores
-            // so we need to drop ${leadTimeInDays} elements from the list of openingTimes
-            .drop(getLeadTimeInDays(venue.openingTimes))
+          ) toMap
         case Left(venueOpeningTimesLookupError) =>
           error(
```
kenoir, line 163: If you move this log into the `venuesOpeningTimesLookup` function you can use the right-biased-ness of Eithers to avoid dealing with the left (you can just map over the Either and ignore dealing with the left as long as you can coerce the Either type to a Map i.e. `.toSeq.toMap`).

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -57,142 +103,123 @@ class SierraItemUpdater(
...
-      .map {
-        case Right(venue) =>
-          venue.openingTimes
+      venuesOpeningTimes = venues match {
+        case Right(venues) =>
+          venues
             .map(
-              openClose => AvailabilitySlot(openClose.open, openClose.close)
+              venue =>
+                venue.title -> venue.openingTimes.map(
```
kenoir, line 157: This is a bit nested, maybe move dealing with the individual venue into its own private function?

items/src/main/scala/weco/api/items/services/VenuesOpeningTimesLookup.scala
```diff
@@ -0,0 +1,71 @@
...
+    Either[VenueOpeningTimesLookupError, List[ContentApiVenue]]
+  ] = {
+    val path = Path(s"venues")
+    val params = Map("title" -> "library", "title" -> title)
+
+    val httpResult = for {
+      response <- client.get(path = path, params = params)
+
+      result <- response.status match {
+        case StatusCodes.OK =>
```
kenoir, line 46: Nitpicky: you could move this into its own function and un-nest things a bit.

items/src/test/scala/weco/api/items/services/SierraItemUpdaterTest.scala
```diff
@@ -0,0 +1,503 @@
...
+        availableDates
+      ),
+      (
+        missingItemResponse(workWithAvailableItemNumber),
+        Seq((contentApiVenueRequest("library"), contentApiVenueResponse())),
+        List(availableItem(workWithAvailableItemNumber)),
+        availableOnlineAccessCondition,
+        availableDates
+      )
+    )
```
kenoir, line 247: Should we have some test cases for deepstore so we test the logic in `deepstoreItemAvailabilities`?

kenoir (commented):
Sorry for the drive-by review! We'll miss each other next week when i'm away and it would have been better to talk through this in person.

The logic here seems sound, and the only thing that might block merging this PR is making sure the logic around deepstore is tested. The comments about unwrapping futures and for comprehensions might be good to look at and talk through with @paul-butcher to see if they do make sense but should not block merge.

No need to wait for my +1 to merge if Paul approves.

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -22,177 +23,190 @@ import scala.concurrent.{ExecutionContext, Future}
...
-    // there is only ever one location per physicalItem and one accessCondition per location,
-    // but this may sometimes be empty if it could not be fetched.
-    val accessCondition = item.locations.headOption
-      .flatMap(_.accessConditions.headOption)
-
-    if (accessCondition.exists(_.isRequestable)) {
-      getAvailableDates(accessCondition.head)
-        .map(availableDates => item.copy(availableDates = Some(availableDates)))
+    item: DisplayItem,
+    sierraItemLocation: Option[SierraLocation]
```
kenoir, line 115: I think you might be able to simplify by checking for a `Some` and unwrapping the option in the calling function, not blocking for merge though.

kenoir (approved):
Nice fixes :)

kenoir:
Part of https://github.com/wellcomecollection/catalogue-api/issues/787

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
agnesgaroux: I pushed it down because the calling function `updateItem` ended up with multiple nested match cases and I thought it looked a bit tangled

PR #798, opened 2024-08-01
lowercase the venue title to create the Map key

agnesgaroux:
`key not found: library` 500 thrown in staging in the previous commit
Due to content-api returning the venue title as "Library" which is used as the key in a Map where we then look for "library"

Lowercasing the title so the look up find a match


kenoir (approved):
🎯

PR #799, opened 2024-08-06
compare dates without time + better tests data

agnesgaroux:
A quick round of testing showed a duplicate date in the dropdown
That's because we were keeping any library opening day that were `>`  deepstore's delivery date.
Since the library opens later than deepstore's working hours, the same day was technically greater, only by a few hours
We now compare only the days, without the hours
Test data improved to reflect this 

Screenshot 2024-08-06 at 13 34 22

NOTE: the list is shorter than it should be. This will be addressed in a separate content-api PR


paul-butcher (approved):
🕰️

PR #801, opened 2024-08-07
compare dates without time + better tests data -take 2

agnesgaroux:
## Pull request checklist

This was merged https://github.com/wellcomecollection/catalogue-api/pull/799
then reverted https://github.com/wellcomecollection/catalogue-api/pull/800
because #799 deployment had a wobble that caused the staging api to go down for a little while

This didn't have anything to do with the code, and the subsequent deployments went without a hitch; we can merge this code back in


paul-butcher (approved):

items/src/main/scala/weco/api/items/services/SierraItemUpdater.scala
```diff
@@ -209,4 +210,11 @@ class SierraItemUpdater(
         } yield updatedItems
     }
   }
+
+  private def parseISOStringToLocalDate(isoString: String): LocalDate =
```
paul-butcher, line 214: Self documenting code makes me smile.

PR #802, opened 2024-08-15
Update API local running

kenoir:
## What does this change?

This change updates the search API to enable running locally. This needed doing to accommodate the new SSO method of developer sign-in and includes a small improvement to allow setting of the pipeline date in dev without a code change that might accidentally get committed. 

## How to test

- [ ] Follow the updated docs, can you run the search API locally?

Easier for developers to run the search API locally.

This change updates configuration to re-point the pipeline to a new index in some cases, however it should only work when the dev environment is detected, and has been tested to ensure it falls back to the hard-coded value when no environment variable is specified.


StepanBrychta (approved):
Apart from the issue with downloading dependencies from S3 (discussed [here](https://wellcome.slack.com/archives/C02ANCYL90E/p1723735928975299)), this works for me.

I'm not sure if this is specific to my environment, but
```
sbt "project search" ~reStart
```

doesn't work for me. It only works when I put ~reStart in quotes:
```
sbt "project search" "~reStart"
```

Assuming this is not just a quirk of my environment/shell, it might be good to update the command in the docs to include the quotes.

PR #804, opened 2024-08-19
run auto formating script with github action

agnesgaroux:
## Pull request checklist

### DIFF
--- a/.github/workflows/autoformat.yml
+++ b/.github/workflows/autoformat.yml
@@ -1,0 +1,12 @@
+# Runs auto-formatting script on push to any branch except main
+name: "Run auto formatting"
+
+on: push
+
+jobs:
+  autoformat:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        id: autoformat
+      - run: bash ./builds/run_formatting.sh


.github/workflows/autoformat.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: eu-west-1
+          role-to-assume: ${{ secrets.GHA_PLATFORM_ECR_READ_ROLE_ARN }}
+      - uses: aws-actions/amazon-ecr-login@v2
+        env:
+          AWS_REGION: eu-west-1
```
kenoir, line 22:
```suggestion
          AWS_REGION: us-east-1
```

kenoir: I think you may need to authorize in this region, see https://github.com/aws-actions/amazon-ecr-login?tab=readme-ov-file#configure-credentials.

This is what the working workflow in wellcomecollection.org does too: https://github.com/wellcomecollection/wellcomecollection.org/blob/main/.github/workflows/prismic-linting.yml#L21C1-L21C32

kenoir (commented):
Will the auto-commit here trigger another unnecessary buildkite build? See comments on https://github.com/wellcomecollection/.github/pull/7

kenoir (approved):
The PR description could do with a bit more context and links to the rest of this work, but this change looks sound. 👍

PR #823, opened 2024-10-28
needs higher version of python

agnesgaroux:
## What does this change?

Following [this](https://github.com/wellcomecollection/catalogue-api/pull/822) `pip install -r requirements.txt` is failing in CI because the new version of the packages are not compatible with python3.7
Tested it locally with python3.12 and it seems to fix the problem 

### Checklist

## How to test

Follow the instructions in catalogue-api/diff_tool/README.md but make sure you're using python3.12
Run ./diff_tool.py --console so as to see the output

diff_tools can run in CI

Might not work out


kenoir (approved):
Yeah, 3.7 is pretty old. Good catch.

PR #829, opened 2024-11-14
Use features vector for kNN queries #5823

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/5823

Depends on https://github.com/wellcomecollection/catalogue-pipeline/pull/2762

Changes the catalogue API to use the full `features` vector instead of the `reducedFeatures` vector when making kNN queries to get visually similar images.

## How to test

This change can be effectively tested once the new ES indexes are in place and a full reindex of all images has been done.

This change is a result of cleaning up the images index by consolidating all the indexed `features` vectors into just one vector. This was done mostly for readability and maintainability reasons rather than to benefit end users. 

Using the full `features` vector instead of the `reducedFeatures` vector should produce better results when recommending visually similar images, but the extent of these improvements is unclear. 

There should be no risks — from the point of view of the catalogue API, this change only involves using a different field name when running kNN queries. The interface of the API remains the same.


common/search/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.3"
 
 services:
   elasticsearch:
-    image: "docker.elastic.co/elasticsearch/elasticsearch:8.4.0"
+    image: "docker.elastic.co/elasticsearch/elasticsearch:8.11.4"
```
kenoir, line 5: 👍

kenoir (approved):
Looks good. This may be open for a bit if we don't do a reindex straight away.

PR #830, opened 2024-11-18
Implement contributor & subject concept filters

kenoir:
## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5654

Follows https://github.com/wellcomecollection/catalogue-pipeline/pull/2746

Resolves https://github.com/wellcomecollection/platform/issues/5821

This change adds the filters:

`source.contributors.concepts`, `source.subjects.concepts` to the `/images` API
`contributors.concepts`, `subjects.concepts` to the `/works` API

In order that concept pages can be populated using concept ids, instead of labels.

### Checklist

- [ ] Does the PR need tests?
        Yes, this PR needs to be updated to add them.
        Yes, incoming an another PR!
        Yes, this PR needs to be updated to add them.

## How to test

- [ ] Run the tests, do they pass?
- [ ] Run locally against the current pipeline, do the filters work?

Better concept pages!

This change does not modify existing API endpoints so should be minimally risky.


agnesgaroux:
Why do some of the filters in the RequestBuilders have this if empty checks?

kenoir:
> Why do some of the filters in the RequestBuilders have this if empty checks?

Yeah, good question. Best answer at the moment is cause I copy pasted the `genre` one. I'll see if I really need them.

kenoir:
> > Why do some of the filters in the RequestBuilders have this if empty checks?
> 
> Yeah, good question. Best answer at the moment is cause I copy pasted the `genre` one. I'll see if I really need them.

@agnesgaroux `NoopQuery` seems to have been used to allow matches of filters where lists were passed to remove empty lists. That can be moved into the case statement and the overridden builder functions filtering out this type removed.

agnesgaroux (approved):

PR #834, opened 2024-12-06
the 1st AvailabilitySlot should be a library open day

agnesgaroux:
## What does this change?

While checking the pickup dates dropdown for the holiday closure, it became apparent that this logic is flawed: if the library is closed on the `firstDeepstoreAvailabilitySlot` the 1st date will be incorrect
Instead, start the list from the 1st library open day after the Deepstore lead time of 10 Deepstore working days

### Checklist

## How to test

Run the tests, do they pass?

LE&E will not get in touch with issues regarding online requesting of offsite items 

There is still a human step in this process (see above) so the risk of incorrect dates is mitigated by the fact that they won't effect the request if it's wrong and get in touch with the user directly


kenoir (approved):
Approving, but with one suggestion: Can we have a test for this, that covers the case specifically where the library is shut but deepstore is not?

PR #838, opened 2025-01-22
Fix labelled ID aggregations bug

StepanBrychta:
## What does this change?

Fixes a bug which causes a 'labelled id' filter to display an id instead of a label in the frontend. See [this thread](https://wellcome.slack.com/archives/C3TQSF63C/p1737477925864379) for more info.

This bug occurs when a query returning 0 results is combined with an active 'labelled id' filter (such as the languages filter or the format filter). Under these circumstances, the unfiltered `selfNested` aggregation works with 0 results and therefore cannot return label buckets for the active filter(s). 

The fix involves moving the unfiltered `selfNested` aggregation into a separate global aggregation to ignore the query and therefore always correctly return the label(s) for the filtered id(s). 

### Checklist

## How to test

This can be tested by locally running the frontend and the catalogue-api and configuring the frontend to connect to the local catalogue-api. 

Additionally, a few new automated tests were added to cover the case of a query returning 0 results while a filter is applied to prevent regressions.

Active filters should always surface the filter's label to the user instead of its id, regardless of how many results are returned or which other filters are active. 

With this fix, the number of aggregations and subaggregations remains the same but one of the subaggregations is moved into its own global aggregation, which might have performance implications. However, I have not observed noticeably slower response times in my testing.

### DIFF
--- a/search/src/test/scala/weco/api/search/services/AggregationsTest.scala
+++ b/search/src/test/scala/weco/api/search/services/AggregationsTest.scala
@@ -190,6 +190,31 @@ class AggregationsTest
           .foreach(subjects => subjects should contain("6rJpSUKd2d"))
       }
     }
+
+    it("correctly returns aggregation labels even when no search results are returned") {
+      withLocalWorksIndex { index =>
+        indexTestDocuments(index, works: _*)
+
+        val searchOptions = createWorksSearchOptionsWith(
+           searchQuery = Option(SearchQuery("awdawdawdawda")), // A search query returning 0 results
+          aggregations =
+            List(WorkAggregationRequest.Format, WorkAggregationRequest.SubjectLabel),
+          filters = List(
+            FormatFilter(List("a")),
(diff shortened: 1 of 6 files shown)


search/src/test/scala/weco/api/search/services/AggregationsTest.scala
```diff
@@ -190,6 +190,31 @@ class AggregationsTest
...
+        val searchOptions = createWorksSearchOptionsWith(
+           searchQuery = Option(SearchQuery("awdawdawdawda")), // A search query returning 0 results
+          aggregations =
+            List(WorkAggregationRequest.Format, WorkAggregationRequest.SubjectLabel),
+          filters = List(
+            FormatFilter(List("a")),
+            SubjectLabelFilter(Seq("fIbfVPkqaf"))
+          )
+        )
+        println(searchOptions.aggregations)
```
kenoir, line 207: not blocking, but stray println?

kenoir (approved):

PR #840, opened 2025-02-07
Update running locally instructions

kenoir:
## What does this change?

This change adds some instructions for running the catalogue API to make it easier for newcomers to the project to get started and help with running locally.

See also: https://github.com/wellcomecollection/wellcomecollection.org/pull/11551

## How to test

- Try and follow the instructions, do they work as expected?

Easier to run things locally, easier development!

Risks should be minimal, this is just a docs change.

### DIFF
--- a/docs/developers.md
+++ b/docs/developers.md
@@ -21,6 +21,23 @@ The CI flow looks as follows:
 - Docker
 - Make
 
+### Installing dependencies
+
+We suggest using [SDKMAN](https://sdkman.io/) to manage Java versions. It can be installed with:
+
+```bash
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
Looks like what we did yesterday indeed, only difference is I ran `11.0.26-amzn`, but feel like it doesn't change much?

gestchild (approved):
Just followed the instructions successfully

kenoir:
> Looks like what we did yesterday indeed, only difference is I ran `11.0.26-amzn`, but feel like it doesn't change much?

I used the version in my environment, I don't think this has any impact on whether it'll work or not.

PR #851, opened 2025-04-01
Snapshot cache invalidation

StepanBrychta:
### Checklist

### DIFF
--- a/snapshots/snapshot_recorder/src/snapshot_recorder.py
+++ b/snapshots/snapshot_recorder/src/snapshot_recorder.py
@@ -8,6 +8,7 @@ Elasticsearch.
 
 import json
 import os
+import time
 
 import boto3
 from elasticsearch import Elasticsearch
@@ -18,12 +19,7 @@ def get_sns_messages(event):
     """
     Extracts messages from an SNS event sent to a Lambda function.
     """
-    try:
-        event_records = event["Records"]
-    except KeyError:
-        pass
-
-    for record in event_records:
+    for record in event["Records"]:
         print(f"Working on record: {record!r}")
         assert record["EventSource"] == "aws:sns", record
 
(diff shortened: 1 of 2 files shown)


snapshots/snapshot_recorder/src/snapshot_recorder.py
```diff
@@ -18,12 +19,7 @@ def get_sns_messages(event):
     """
     Extracts messages from an SNS event sent to a Lambda function.
     """
-    try:
-        event_records = event["Records"]
```
agnesgaroux, line 22: Did you remove this check because there's virtually no way SNS will come without `Records`?

snapshots/snapshot_recorder/src/snapshot_recorder.py
```diff
@@ -63,6 +59,26 @@ def prepare_message_for_indexing(completed_snapshot):
     }
 
 
+def invalidate_old_snapshot(completed_snapshot):
+    """
+    Removes any cached versions of the snapshot from the CloudFront distribution, ensuring
+    that the next time someone accesses the snapshot URL, they receive the latest version.
+    """
+    s3_key = completed_snapshot["snapshotResult"]["s3Location"]["key"]
```
agnesgaroux, line 67: Will this evaluate to `catalogue/v2/`, meaning the cached image snapshot will be invalidated too?

snapshots/terraform/stack/snapshot_recorder/main.tf
```diff
@@ -1,5 +1,6 @@
 locals {
-  secret_id = "catalogue/snapshots/write_user"
+  secret_id                       = "catalogue/snapshots/write_user"
+  collection_data_distribution_id = "E26LQLWZB3P1Q0"
```
agnesgaroux, line 3: Should we output the distribution id [here](https://github.com/wellcomecollection/catalogue-api/blob/main/terraform/snapshots/outputs.tf) and get it from the remote state?

snapshots/snapshot_recorder/src/snapshot_recorder.py
StepanBrychta: This will evaluate to the full key of the snapshot, so `catalogue/v2/works.json.gz` for works and `/catalogue/v2/images.json.gz` for images.

I think we are running the snapshot recorder separately for the two snapshots, so the keys will also be invalidated separately as part of their respective run.

snapshots/snapshot_recorder/src/snapshot_recorder.py
StepanBrychta: The way the error handling was implemented, it was not preventing an exception from occurring.

In a scenario where "Records" does not exist as a key in `event`, intercepting the `KeyError` exception would cause a different exception to be raised on the next line since `event_records` would be undefined.

agnesgaroux (approved):

PR #872, opened 2025-11-25
Switch to new works index

StepanBrychta:
## What does this change?

Switch to a new works index (2025-10-09) within the production cluster. This index is populated using the graph pipeline.

### Checklist

## How to test

Run the frontend locally connected to the index. Do work pages look as expected?

The new graph pipeline is used to populate the production works index. There should be no differences in the contents of the new index and the current production index (except for a few deliberate changes, including the removal of the `succeededBy` and `preceededBy` fields in work hierarchies, and some data quality improvements).

This is a major change which involves switching to a new set of services for populating the final works index (retiring the Scala relation embedder and works ingestor).

However, risks are relatively small, since we did a full comparison of the two indexes (see [here](https://wellcome.slack.com/archives/C02ANCYL90E/p1763554599296169)) and did not find any unexpected differences. 

If issues do appear, we have the option of switching back to the old index (2025-10-02).

### DIFF
--- a/common/search/src/main/scala/weco/api/search/models/ElasticConfig.scala
+++ b/common/search/src/main/scala/weco/api/search/models/ElasticConfig.scala
@@ -12,7 +12,7 @@ object ElasticConfig {
   // We use this to share config across Scala API applications
   // i.e. The API and the snapshot generator.
   val pipelineDate = "2025-10-02"
-  val indexDateWorks = "2025-11-20"
+  val indexDateWorks = "2025-10-09"
   val indexDateImages = "2025-10-02"
 }
 


common/search/src/main/scala/weco/api/search/models/ElasticConfig.scala
```diff
@@ -12,17 +12,27 @@ object ElasticConfig {
   // We use this to share config across Scala API applications
   // i.e. The API and the snapshot generator.
   val pipelineDate = "2025-10-02"
+  val indexDateWorks = "2025-11-20"
```
kenoir: 2025-10-09

kenoir (approved):

PR #901, opened 2026-02-09
Multi-cluster search api with semantic search

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/catalogue-api/issues/897

* Add support for additional/secondary Elasticsearch clusters to the search API:
    * Additional clusters can be configured in `application.conf`
    * Separate `ClusterConfig` and ResilientElasticClient objects are created for each additional cluster
    * All requests specifying an `elasticCluster` query parameter are routed to the corresponding cluster
* Add support for semantic search against the works index:
    * Semantic search is disabled by default
    * To enable semantic search, associate a `SemanticConfig` object with a given `ClusterConfig`
    * Both sparse vector and dense vector (knn) queries are supported

## How to test

The PR comes with two pre-configured `multiCluster` config blocks — `elser` and `openai` (both with semantic search enabled). To test:
1. Run the catalogue API locally
2. Send a GET request with the `elasticCluster` parameter, e.g. `/works?elasticCluster=someClusterLabel`
3. Send a GET request without the `elasticCluster` parameter (to verify backwards compatibility)

* The search API sends any requests including a `elasticCluster` query parameter to additional Elasticsearch clusters (if configured)
* Lexical queries are enriched with semantic search via RRF (if enabled)
* Full backwards compatibility: By default, all requests are routed to the production pipeline cluster with semantic search disabled

* Backwards compatibility: This change should be fully backwards compatible (to be verified by unit tests, end to end tests in staging, and rank tests)
* Server overload: Experimental semantic search functionality runs on the same server as production functionality, which might overload the server. However, expected semantic search traffic is very low (internal only) and our ECS deployment is self-healing.
* Exposed indexes: New semantic search indexes (containing ELSER and OpenAI embeddings), though not technically in production, are exposed via our public API


search/src/main/resources/application.conf
```diff
@@ -16,20 +16,16 @@ aws.metrics.namespace=${?metrics_namespace}
...
   semanticVectorType = "sparse"
 }
 
 multiCluster.openai {
-  customHost = "semantic-playground-b28f61.es.eu-west-1.aws.elastic.cloud"
-  customPort = 443
-  customProtocol = "https"
-  customApiKeySecretPath = "agnes/elasticsearch/semantic-playground"
+  hostSecretPath = "agnes/elasticsearch/semantic-playground-host"
+  apiKeySecretPath = "agnes/elasticsearch/semantic-playground"
```
agnesgaroux: I think we can call them secrets something else now 😂

snapshots/terraform/terraform.tf
```diff
@@ -1,6 +1,8 @@
 terraform {
```
agnesgaroux: Pushed this tf by mistake but it should be like that anyway. Nothing to apply

kenoir (approved):

PR #905, opened 2026-02-19
Items.locations.created date filter

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/wellcomecollection.org/issues/12766

Adding a filter on `items.locations.createdDate` so as to exclude works that have been too recently digitised. This aims to allow time for the thumbnails to be generated and prevent failing e2e test when the thumbnail URLs return 404

### Checklist

- [x] Do you need to update the [Catalogue API Swagger][swagger]?

## How to test

Run tests

Run `sbt "project search" "~reStart"` from the root of the project
http://localhost:8080/works?sortOrder=desc&sort=items.locations.createdDate
-> first result is `u3epmef4` from "2026-02-18T08:57:26Z"
http://localhost:8080/works?sortOrder=desc&sort=items.locations.createdDate&items.locations.createdDate.to=2025-12-31
-> first result is `u2rpht4h` from "2025-12-18T20:17:16Z"

No works are displayed on the new online page that do not have a thumbnail 

This filter will only be used behind the scenes to fetch content for the new online page, it will not be user facing

### DIFF
--- a/search/src/main/scala/weco/api/search/models/DocumentFilter.scala
+++ b/search/src/main/scala/weco/api/search/models/DocumentFilter.scala
@@ -23,6 +23,11 @@ case class DateRangeFilter(
 ) extends WorkFilter
     with ImageFilter
 
+case class ItemsLocationsCreatedDateFilter(
+  fromDate: Option[LocalDate],
+  toDate: Option[LocalDate]
+) extends WorkFilter
+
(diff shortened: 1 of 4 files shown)


kenoir (approved):
Looks good - the Swagger needs updating I think.

PR #923, opened 2026-06-03
Tweak tests for compatibility with Python pipeline

StepanBrychta:
## What does this change?

Following the migration of work/image document generators to Python (see https://github.com/wellcomecollection/catalogue-pipeline/pull/3402), this PR makes the catalogue API tests compatible with the new documents:
* Modify the `copy_test_documents.py` script to copy documents from the new location in the catalogue pipeline repository
* Change hardcoded ID/title values in various tests to match values in the new documents
* Create Python scripts to automate the generation of expected responses for WorksIncludesTest and ImagesIncludesTest.
* Remove a few outdated tests (e.g. work documents no longer include `precededBy` and `succeededBy` fields).

## How to test

Run the tests and make sure they all pass.

Catalogue API tests are now compatible with the Python generated documents (and the Python ingestor more broadly).


search/src/test/scala/weco/api/search/images/ImagesFiltersTest.scala
```diff
@@ -100,17 +100,6 @@ class ImagesFiltersTest extends AnyFunSpec with ApiImagesTestBase {
...
-      withGenreFilterRecords { routes =>
-        assertJsonResponse(
-          routes,
-          path = s"$rootPath/images?source.genres.label=Dodo%20divination"
-        ) {
-          Status.OK -> emptyJsonResult
-        }
-      }
-    }
-
```
StepanBrychta, line 113: In the original test documents, there was no work with the `Dodo divination` subject due to a bug in the Scala image generators. The test succeeded, but not for the reason the author intended, so I'm removing it.

search/src/test/scala/weco/api/search/works/WorksIncludesTest.scala
StepanBrychta, line 1: I extracted the expected responses for this test into JSON files under `resources/expected_responses`, which are now generated automatically using a Python script.

kenoir (approved):

PR #936, opened 2026-07-07
Add an axiell-collections-testing multiCluster entry for per-request pipeline routing

kenoir:
## What does this change?

Registers the new (Axiell/FOLIO) pipeline's shared ES cluster as a `multiCluster` entry labelled `axiell-collections-testing`, so any works/images request (list and detail) can be routed to the new pipeline with `?elasticCluster=axiell-collections-testing`. This lets staff preview and compare the new pipeline's output before we flip the default, using the existing multi-cluster mechanism. The change is config, tests, and docs only, with no production code changes.

The entry reads its connection secrets from the `elasticsearch/pipeline_storage_2026-07-03/` prefix, written to the catalogue account by the `pipeline_new` stack in the catalogue-pipeline repo.

As with the other additional clusters, the entry is non-essential: if it fails to build at startup it is logged and dropped, and requests selecting it get an explicit 404 rather than falling back to the old pipeline.

The paired front-end change is wellcomecollection/wellcomecollection.org#13253, which adds the staff toggle that sends the param.

Part of wellcomecollection/platform#6421

### Checklist

- [x] Does this patch need a change to the documentation? `docs/multi-cluster-setup.md` updated (new entry, `pipeline_storage_2026-07-03` secret paths, local-dev note, default-flip and rollback steps)
- [ ] Do you need to update the [Catalogue API Swagger][swagger]? No, `elasticCluster` is an existing internal-use param

## How to test

`ElasticClusterParamTest` (5 tests) covers routing for `/works`, `/works/{id}`, `/images`, `/images/{id}` on a cluster with both indexes configured.

Verified end-to-end locally (2026-07-21) with the paired front-end PR driving the param through the local nginx `api-dev` setup:

- With the committed config the cluster builds at startup and toggled requests return 200 with results from the `2026-07-03` indexes, which are empty until the new pipeline populates them. When running the API outside the VPC (e.g. locally), swap the entry's `hostSecretPath` to the `public_host` secret per `docs/multi-cluster-setup.md`.
- If the secrets or cluster are ever missing, the cluster is dropped at startup and `?elasticCluster=axiell-collections-testing` returns an explicit `404 "Cluster 'axiell-collections-testing' is not configured"`, while default traffic is unaffected.

Once the new pipeline populates the indexes, `GET /catalogue/v2/works?query=medicine&elasticCluster=axiell-collections-testing` returns results from `works-indexed-2026-07-03`.

Staff can compare old and new pipeline results on the same environment by adding or removing the param (or using the front-end toggle), supporting sign-off of the Axiell/FOLIO migration without a deploy.

- Requests without the param are untouched; the default cluster wiring is unchanged.
- A misconfigured or unavailable cluster can't crash the API (non-essential build) and can't silently serve old data (explicit 404).
- If the secrets are ever missing, deploying is still safe: the cluster is dropped with a logged error.

### DIFF
--- a/search/src/main/resources/application.conf
+++ b/search/src/main/resources/application.conf
@@ -40,11 +40,11 @@ multiCluster.openai {
 }
 
 # The new (Axiell/FOLIO) pipeline; see docs/multi-cluster-setup.md
-multiCluster.new-pipeline {
-  hostSecretPath = "elasticsearch/es-cluster-2026-07-03/private_host"
-  portSecretPath = "elasticsearch/es-cluster-2026-07-03/port"
-  protocolSecretPath = "elasticsearch/es-cluster-2026-07-03/protocol"
-  apiKeySecretPath = "elasticsearch/es-cluster-2026-07-03/catalogue_api/api_key"
+multiCluster.axiell-collections-testing {
+  hostSecretPath = "elasticsearch/pipeline_storage_2026-07-03/private_host"
+  portSecretPath = "elasticsearch/pipeline_storage_2026-07-03/port"
+  protocolSecretPath = "elasticsearch/pipeline_storage_2026-07-03/protocol"
+  apiKeySecretPath = "elasticsearch/pipeline_storage_2026-07-03/catalogue_api/api_key"
   worksIndex = "works-indexed-2026-07-03"
   imagesIndex = "images-indexed-2026-07-03"
 }
--- a/search/src/main/resources/application.conf
+++ b/search/src/main/resources/application.conf
@@ -38,3 +38,13 @@ multiCluster.openai {
     numCandidates = 500
   }
 }
+
+# The new (Axiell/FOLIO) pipeline; see docs/multi-cluster-setup.md
+multiCluster.new-pipeline {
+  hostSecretPath = "elasticsearch/es_cluster_2026-07-03/private_host"
+  portSecretPath = "elasticsearch/es_cluster_2026-07-03/port"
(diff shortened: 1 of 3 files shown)


search/src/main/resources/application.conf
```diff
@@ -38,3 +38,13 @@ multiCluster.openai {
     numCandidates = 500
   }
 }
+
+# The new (Axiell/FOLIO) pipeline; see docs/multi-cluster-setup.md
+multiCluster.new-pipeline {
+  hostSecretPath = "elasticsearch/es_cluster_2026-07-03/private_host"
+  portSecretPath = "elasticsearch/es_cluster_2026-07-03/port"
+  protocolSecretPath = "elasticsearch/es_cluster_2026-07-03/protocol"
+  apiKeySecretPath = "elasticsearch/pipeline_storage_2026-07-03/catalogue_api/api_key"
```
agnesgaroux, line 47: All the secrets should have the same prefix `elasticsearch/es-cluster-2026-07-03/` (all hyphens)

kenoir: Good catch. Fixed in 9835b00e - all four secret paths now use the elasticsearch/es-cluster-2026-07-03/ prefix (hyphens throughout), including the api_key path which had also picked up the pipeline_storage_ prefix. Updated the matching public_host reference in the doc too.

agnesgaroux: Belatedly realised that this solution was not workable
Reverted to storing secrets where consumers expect them, ie. elasticsearch/pipeline_storage_<pipeline_date>

StepanBrychta (approved):

PR #955, opened 2026-08-12
Assert every documented response field appears in a fixture

kenoir:
## What does this change?

`OpenApiSpecResponseTest` caught fields the pipeline emits that the spec misses, but nothing caught the reverse: a field removed from the display model stayed documented forever unless a schema happened to mark it required. This is step 2 of wellcomecollection/platform#6514.

The new assertion enumerates every property of every component reachable from the `Work` and `Image` schemas and requires each to appear in at least one fixture. Comparison is per component rather than per position, so `Identifier.value` counts as exercised wherever an identifier appears and a rarely-identified nested entity doesn't need its own fixture. A `oneOf` with a discriminator credits a field only to the branch the document names, so a field shared between location types can't be masked by the sibling branch. Guard tests back it: an invented schema property must be reported missing, an allowlisted field that becomes exercised fails, and the spec is pinned to the schema keywords the walks understand (an `allOf` refactor would otherwise silently exempt its subtree from every check in the file).

The first run surfaced 18 documented fields no fixture contained, and this PR resolves all of them:

- `precededBy`, `succeededBy` and nested `RelatedWork.partOf` were lost in the pipeline's move to catalogue_graph (absent from the Pydantic display models, confirmed against the live API), so they leave the spec. The include values are accepted and ignored rather than rejected, because iiif-builder sends them on every request; an enum test pins them as undocumented, and they can be rejected once iiif-builder stops. The response keys are stripped unconditionally so a rollback or `elasticCluster` override serving an old index can't leak them.
- Ten were real fields the deterministic fixture sample never populated: `Work.physicalDescription`, `Work.referenceNumber`, `Work.createdDate`, `Work.currentFrequency`, `Item.title`, `Item.note`, `AccessCondition.note`, `AccessCondition.terms`, `ProductionEvent.function` and `Agent.identifiers`. wellcomecollection/catalogue-pipeline#3566 teaches the generators to populate them via a new `work.visible.rare-fields.json` fixture, which this PR pulls in.
- Ids and identifiers on `Period` and `Place` leave the spec too: the ingestor's display transformer drops them even for identified concepts, and if it ever stops, the existing undocumented-fields walk forces re-documentation on the next fixture refresh.
- One field stays allowlisted with its reason in the code: `Image.withSimilarFeatures` is spliced in by the API at request time, so no indexed fixture can carry it.

Also fixed along the way, prompted by review: the stress tests passed `includes=` (a parameter name the API ignores) so they exercised no includes at all, now corrected to `include=`; and `copy_test_documents.py` now mirrors upstream deletions instead of only copying.

This branch carries the fixture wellcomecollection/catalogue-pipeline#3566 generates; that PR merged on 2026-08-17, so the fixtures here now match pipeline main and this is ready for review.

### Checklist

- [x] Does this patch need a change to the documentation? The test table in `reference/README.md` is updated; the developers site picks up the spec change via the existing sync. Its prose examples still mention the removed includes and need a small follow-up there.
- [x] Does this change the API surface? Yes: the spec drops the dead response fields. The deprecated include values remain accepted (and ignored) so no client breaks.

## How to test

```
yarn lint:openapi
sbt "search/testOnly weco.api.search.openapi.*"
sbt search/test
```

All 406 search tests pass. To see the new assertion work, add an invented property to the `Work` schema in `reference/catalogue.yaml` and `OpenApiSpecResponseTest` fails naming it.

A display-model removal now fails CI the moment fixtures refresh, instead of leaving the spec documenting fields that no longer exist. The 18 stale documented fields found on the first run drop to 1, with its reason recorded.

- iiif-builder sends the deprecated includes by default, which is why they are tolerated rather than rejected; nothing changes for it. Rejecting them later needs an iiif-builder release first.
- Old indices still contain the removed fields, so the response path strips those keys unconditionally rather than relying on include gating.
- The remaining allowlist entry could hide a future regression, which is why the guard test fails if it becomes exercised.

### DIFF
--- a/common/search/src/test/resources/test_documents/work.visible.rare-fields.json
+++ b/common/search/src/test/resources/test_documents/work.visible.rare-fields.json
@@ -1,0 +1,241 @@
+{
+  "description": "a work with every rarely-populated display field",
+  "id": "4cbgoe8p",
+  "document": {
+    "type": "Visible",
+    "debug": {
+      "source": {
+        "id": "4cbgoe8p",
+        "identifier": {
+          "identifierType": {
+            "id": "sierra-system-number"
+          },
+          "ontologyType": "Work",
+          "value": "3QF59FdNQJ"
+        },
+        "version": 1,
+        "modifiedTime": "2025-01-01T00:00:00Z"
+      },
+      "mergedTime": "2025-01-01T00:00:00Z",
+      "indexedTime": "2001-01-01T01:01:01Z",
+      "mergeCandidates": [],
+      "redirectSources": []
+    },
+    "query": {
+      "id": "4cbgoe8p",
+      "title": "A work with every rarely-populated display field",
+      "referenceNumber": "WA/HMM/BU/1",
+      "physicalDescription": "1 volume ; 23 cm",
+      "alternativeTitles": [],
+      "languages.label": [],
+      "sourceIdentifier.value": "3QF59FdNQJ",
+      "identifiers.value": ["4cbgoe8p", "3QF59FdNQJ"],
+      "images.id": [],
+      "images.identifiers.value": [],
+      "items.identifiers.value": ["Y14y2h0Rzl"],
+      "items.id": ["l5mem3bu"],
+      "items.shelfmark.value": ["Shelfmark: NaTDHzrYbX"],
+      "notes.contents": [],
+      "partOf.title": [],
+      "production.label": ["London", "Printed for the author", "1897"],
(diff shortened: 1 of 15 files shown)


agnesgaroux (approved):
Medium: `copy_test_documents.py` is a delete-then-copy with no guard, so a stale pipeline clone or a wrong path produces a confusing test failure or wipes all 238 fixtures with exit 0. Collecting source paths first and exiting when the list is empty fixes both.

